Issue #21, opened 2022-03-29
Update CONTRIBUTING.md with installation instructions for Red Hat RPM based distributions


nikitabugrovsky:
**Is your feature request related to a problem? Please describe.**

At the moment installation instructions reflect steps required to run builds on Fedora distribution. Steps for Red Hat RPM based distributions are slightly different. It is worth mentioning it.

**Describe the solution you'd like**

Extend CONTRIBUTING.md with steps relevant for Red Hat RPM based distributions.

jesperpedersen:
First, the PostgreSQL ecosystem has discontinued support for CentOS 8 due to its unstable ABI.

Second, there are additional RPM based distributions, so I think it would be better to just write

```
For Red Hat RPM based distributions ...
```

nikitabugrovsky:
@jesperpedersen thanks a lot for your comments. I made changes to the PR accordingly.

PR #22, opened 2022-03-29
 Installation instructions for Red Hat RPM based distributions

nikitabugrovsky:
Issue: pgexporter/pgexporter#21

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,3 +3,4 @@ pgexporter was created by the following authors:
 Jesper Pedersen <jesper.pedersen@redhat.com>
 David Fetter <david@fetter.org>
 Will Leinweber <will@bitfission.com>
+Nikita Bugrovsky <nbugrovs@redhat.com>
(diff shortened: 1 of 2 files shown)


jesperpedersen:
You need to add yourself to the `AUTHORS` file as well

jesperpedersen:
There is no reason to have the same command twice...

nikitabugrovsky:
@jesperpedersen with e45ab3b to 2a3fe6e I changed commit message as well.

jesperpedersen:
You need to remove the "Fedora" part still... do

```
[#21] Installation instructions for Red Hat RPM based distributions
```

as the commit message - you don't need to list the files changed.

nikitabugrovsky:
@jesperpedersen there is one difference in package naming between Fedora & RHEL based: `liblz4`, `liblz4-dev` vs `lz4`,` lz4-devel` respectively. Do you think it would be better to stick to 2 different lines?

jesperpedersen:
Ah, those are typos -- `pgexporter` doesn't depend on LZ4

jesperpedersen:
Merged.

Issue #23, opened 2022-03-30
Return 400 Bad Request for any method except 'GET'


nikitabugrovsky:
**Is your feature request related to a problem? Please describe.**

At the moment if we send request to pgexporter endpoint using any method except 'GET',  we receive 403 Forbidden. 403 indicates that the server understands the request but refuses to authorize. That is not exactly the case with the exporter.

**Describe the solution you'd like**

Introduce new check in the code that will allow to return 400 Bad Request for any method except 'GET'.

Ref: [Hypertext Transfer Protocol -- HTTP/1.1](https://www.w3.org/Protocols/rfc2616/rfc2616.html)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)

PR #24, opened 2022-03-30
Return 400 Bad Request for any method except 'GET'

nikitabugrovsky:
Issue: pgexporter/pgexporter#23

### DIFF
--- a/src/libpgexporter/prometheus.c
+++ b/src/libpgexporter/prometheus.c
@@ -49,11 +49,13 @@
 #define PAGE_UNKNOWN 0
 #define PAGE_HOME    1
 #define PAGE_METRICS 2
+#define BAD_REQUEST  3
 
 static int resolve_page(struct message* msg);
 static int unknown_page(int client_fd);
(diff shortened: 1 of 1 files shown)


src/libpgexporter/prometheus.c
```diff
@@ -138,7 +144,7 @@ resolve_page(struct message* msg)
    if (strncmp((char*)msg->data, "GET", 3) != 0)
```
jesperpedersen, line 144: You need to check that msg->length is valid

src/libpgexporter/prometheus.c
```diff
@@ -138,7 +144,7 @@ resolve_page(struct message* msg)
...
    {
       pgexporter_log_debug("Promethus: Not a GET request");
```
jesperpedersen, line 146: Spelling

jesperpedersen (changes requested):

jesperpedersen:
Please, submit a similar pull request to pgagroal and pgmoneta

Issue #40, opened 2022-09-02
Custom Metrics


xRay2016:
The pgexporter should support the customized metrics.

jesperpedersen:
There are a couple of improvements that could be done as part of this issue.

1) `server` with `both`, `primary` and `replica` as values
2) `type` with `histogram`

We will leave them here for now...

xRay2016:
I will submit the server type filter firstly.

jesperpedersen:
The core functionality for custom queries is in.

Remaining work will be done as separate issues.

This concludes @xRay2016 's Google Summer of Code project ! Big thanks for all your effort !

PR #42, opened 2022-09-07
[#40] yaml options

xRay2016:
Add -Y / --yaml feature too that loads the file/directory from the command line.

If this is defined in command line, the `metrics_path` in config is overrode.


src/main.c
```diff
@@ -196,6 +196,7 @@ usage(void)
    printf("  -c, --config CONFIG_FILE Set the path to the pgexporter.conf file\n");
    printf("  -u, --users USERS_FILE   Set the path to the pgexporter_users.conf file\n");
    printf("  -A, --admins ADMINS_FILE Set the path to the pgexporter_admins.conf file\n");
+   printf("  -Y, --yaml CUSTOM_METRICS_FILE Set the path to the custom_metrics.yaml file\n");
```
jesperpedersen, line 199: Move the other descriptions such that they align, e.g.

```
... Set the path
... Set the path
... Run as a
```

src/main.c
```diff
@@ -431,6 +437,10 @@ main(int argc, char** argv)
    }
 
    config = (struct configuration*)shmem;
+   if (yaml_path!=NULL)
```
jesperpedersen, line 440: `uncrustify` - otherwise ok

src/main.c
jesperpedersen: And it is directory too, right ?

jesperpedersen: Maybe

```
Set the path to a YAML file/directory
```

src/main.c
```diff
@@ -193,12 +193,13 @@ usage(void)
...
-   printf("  -c, --config CONFIG_FILE Set the path to the pgexporter.conf file\n");
-   printf("  -u, --users USERS_FILE   Set the path to the pgexporter_users.conf file\n");
-   printf("  -A, --admins ADMINS_FILE Set the path to the pgexporter_admins.conf file\n");
-   printf("  -d, --daemon             Run as a daemon\n");
-   printf("  -V, --version            Display version information\n");
-   printf("  -?, --help               Display help\n");
+   printf("  -c, --config CONFIG_FILE        Set the path to the pgexporter.conf file\n");
+   printf("  -u, --users USERS_FILE          Set the path to the pgexporter_users.conf file\n");
+   printf("  -A, --admins ADMINS_FILE        Set the path to the pgexporter_admins.conf file\n");
+   printf("  -Y, --yaml CUSTOM_METRICS_FILE  Set the path to the custom_metrics.yaml file\n");
```
jesperpedersen, line 199: Should we do

```
-Y, --yaml METRICS_FILE _DIR Set the path to YAML file/directory\n
```
instead ?

jesperpedersen (changes requested):

src/main.c
jesperpedersen: Likely need realign with the descriptions again

jesperpedersen:
Squashed, rebased and merged.

Issue #32, opened 2022-05-08
Improve built-in /metrics endpoints to use both counter and gauge


nikitabugrovsky:
**Describe the bug**

At the moment `gauge` metric primitive is being hard coded for all exposed metrics:  https://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/prometheus.c#L1450
we need to be able to to differentiate between different metrics primitives  (counters vs gauges) and report metric type correctly.

**To Reproduce**

1. bin/pgexporter -c /etc/pgexporter/pgexporter.conf -u /etc/pgexporter/pgexporter_users.conf
2. make sure /metrics endpoint is up & running
3. See the following line being reported by pgexporter ` #TYPE <some_metric_name> gauge`

**Version**
pgexporter 0.2.2

jesperpedersen:
You will need to go through all the queries and figure out which should be `gauge` and which ones should be `counter`.

I would add this information to https://github.com/pgexporter/pgexporter/blob/main/src/include/queries.h#L51

jesperpedersen:
Custom queries have been merged, so they can be used in-place of the built-in ones

Aditya0049:
To fix this bug, you need to modify the code of the pgexporter to correctly report the metric types of each metric. Here is a possible approach for doing this:

1. Modify the prometheus.c file to define a new MetricType enum that represents the different types of Prometheus metrics. For example:
**c**
```
typedef enum {
  METRIC_TYPE_COUNTER,
  METRIC_TYPE_GAUGE,
  METRIC_TYPE_HISTOGRAM,
  METRIC_TYPE_SUMMARY,
} MetricType;
```

2. Modify the prometheus.c file to define a new metric_t struct that includes the metric name, value, and type. For example:

```
typedef struct metric_s {
  char *name;
  double value;
  MetricType type;
  struct metric_s *next;
} metric_t;
```

3. Modify the prometheus.c file to add a new **type** parameter to the prom_add_metric() function that specifies the type of the metric being added. For example:

void prom_add_metric(char *name, double value, MetricType type) {
  metric_t *m = prom_find_metric(name);

```
  if (m == NULL) {
    m = malloc(sizeof(metric_t));
    m->name = strdup(name);
    m->value = value;
    m->type = type;
    m->next = metrics;
    metrics = m;
  } else {
    m->value = value;
  }
}
```

4. Modify the prometheus.c file to modify the prom_dump_metrics() function to output the correct type for each metric. For example:

```
[...]

PR #62, opened 2023-04-25
[#32] Improve built-in /metrics endpoints to use both counter and gauge

resyfer:
@jesperpedersen This is the implementation of what I had in my mind for #32. I still haven't added an install instruction for `internal.yml` file to have the path `/etc/pgexporter/internal.yml`, but using the following command before running the binary should do it.

```sh
sudo cp ./contrib/yaml/internal.yml /etc/pgexporter/internal.yml
```

I am not much familiar with writing CMake files or Spec files for RPM, so I'll look into those for this when this idea gets a green signal.

PS: This is just a bare-bones working change to the repo to demonstrate my idea. This needs work here and there, which I'll be glad to do.

### DIFF
--- a/src/include/internal.h
+++ b/src/include/internal.h
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2023 Red Hat
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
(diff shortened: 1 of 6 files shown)


jesperpedersen:
Ok, I get where you are coming from.

However, `internal.yml` def lives in `src/yaml` instead, and you can't assume it should loaded from `/etc/pgexporter` as the configuration can be overwritten by `-c`.

Start by investigation if it can be embedded into the shared library, and loaded from there. Or do it as an `internal.h` with the text string.

Latter is likely more acceptable...

As is, no

jesperpedersen:
Much happier with that - go ahead

resyfer:
@jesperpedersen PTAL

resyfer:
I have yet to identify the metrics that are better represented by counter than a gauge and I'll start looking into it.

src/include/internal.h
```diff
@@ -0,0 +1,243 @@
+#ifndef PGEXPORTER_INTERNAL_H
```
jesperpedersen, line 1: Remember header

src/include/pgexporter.h
```diff
@@ -61,12 +61,13 @@ extern "C" {
 #define MAX_USERNAME_LENGTH  128
 #define MAX_PASSWORD_LENGTH 1024
 
-#define MAX_PATH 1024
-#define MISC_LENGTH 128
-#define NUMBER_OF_SERVERS 64
-#define NUMBER_OF_USERS    64
-#define NUMBER_OF_ADMINS    8
-#define NUMBER_OF_METRICS 256
+#define MAX_PATH              1024
```
jesperpedersen, line 64: If you are doing this then right-align

src/include/pgexporter.h
```diff
@@ -84,15 +85,16 @@ extern "C" {
...
 
 #define LABEL_TYPE      0
 #define COUNTER_TYPE    1
 #define GAUGE_TYPE      2
 #define HISTOGRAM_TYPE  3
 
-#define SORT_NAME  0
-#define SORT_DATA0 1
+#define SORT_NAME 0
+#define SORT_DATA 1
```
jesperpedersen, line 97: Lets keep the `SORT_DATA0` name - we may want to sort by multiple data columns in the future

src/libpgexporter/yaml_configuration.c
```diff
@@ -118,14 +138,9 @@ pgexporter_read_yaml(struct prometheus* prometheus, char* filename, int* number_
...
 
-   file = fopen(filename, "r");
-   if (file == NULL)
-   {
-      pgexporter_log_error("pgexporter: fopen error %s", strerror(errno));
-      return 1;
-   }
    if (!yaml_parser_initialize(&parser))
    {
+      printf("Hi\n");
```
jesperpedersen, line 143: Hello !

src/main.c
```diff
@@ -27,18 +27,20 @@
...
 
 /* pgexporter */
-#include <pgexporter.h>
 #include <configuration.h>
-#include <yaml_configuration.h>
+#include <internal.h>
 #include <logging.h>
 #include <management.h>
 #include <network.h>
+#include <pgexporter.h>
```
jesperpedersen, line 35: This one is always first

src/main.c
```diff
@@ -27,18 +27,20 @@
...
 #include <prometheus.h>
 #include <queries.h>
 #include <remote.h>
 #include <security.h>
+#include <stdbool.h>
```
jesperpedersen, line 40: System include

jesperpedersen (changes requested):

jesperpedersen:
Sounds good - I gave the patch a quick scan...

resyfer:
@jesperpedersen I've made the requested changes, and I'll give a PTAL after finishing looking into which metrics need to be collectors

jesperpedersen:
Merged.

resyfer:
@jesperpedersen Thank you for the review and thank you for the (very long) discussion around it! I hope this solves the problem that the issue aimed to point out.

Issue #64, opened 2023-05-13
Retrieving Server's Information 


resyfer:
#63's [comment on `Breakdown`'s Part 1](https://github.com/pgexporter/pgexporter/discussions/63#discussioncomment-5889882)

PR #65, opened 2023-05-13
[#64] Retrieving Server's Information

resyfer:


src/include/internal.h
```diff
@@ -264,6 +264,11 @@ extern "C" {
    "        type: counter\n" \
    "        description: pg_stat_database_conflicts_confl_deadlock"
 
+/**
+ * SQL Query for retrieving floor of server's PostgreSQL version.
+ */
+#define VERSION_SQL "SELECT split_part(split_part(version(), ' ', 2), '.', 1);"
+
```
jesperpedersen, line 271: I think we can put this SQL directly in the function

src/include/pgexporter.h
```diff
@@ -183,6 +185,8 @@ extern void* shmem;
  */
 extern void* prometheus_cache_shmem;
 
+typedef unsigned int version_t;
+
```
jesperpedersen, line 189: I don't think we need a `typedef` for this

src/include/pgexporter.h
```diff
@@ -198,6 +202,7 @@ struct server
    bool new;                           /**< Is the connection new */
    bool extension;                     /**< Is the pgexporter_ext extension installed */
    int state;                          /**< The state of the server */
+   version_t version;
```
jesperpedersen, line 205: `char` should be enough

jesperpedersen (changes requested):

jesperpedersen (changes requested):

src/include/queries.h
```diff
@@ -86,6 +86,17 @@ pgexporter_query_get_functions(int server, struct query** query);
...
 pgexporter_query_execute(int server, char* sql, char* tag, struct query** query);
 
+/**
+
+ * Extracts the server's version as a char (minimum integral data type), and sets
+ * the value of `config->servers[server].version` to it.
+ *
+ * @param server The server index in config
+ * @return Server version as a char data type
+ */
```
jesperpedersen, line 96: We are not really doing that though...

Thinking about it - lets just keep it as a status code - so `int`, and assign internally in the function

jesperpedersen (changes requested):

src/include/queries.h
resyfer: Oh I am really sorry about this. I was travelling back to my place when I made this change and pushed as soon as I got internet, and it was a lapse of judgement on my part.

Yes I am indeed assigning the version internally and the returned value is just 0 if all's ok and 1 if not.

I'll change it back immediately.

jesperpedersen: np

jesperpedersen:
Merged.

resyfer:
Thank you for the review!

PR #68, opened 2023-05-25
Allow building without systemd and execinfo.h

jirutka:
These changes are needed to build pgexporter on Alpine Linux.


resyfer:
If this goes through, it would also enable pgexporter to be shipped with Alpine Linux as a comparatively small Docker Image 🙌

jirutka:
I've already packaged it in Alpine Linux using this patch.

https://pkgs.alpinelinux.org/packages?name=pgexporter

jesperpedersen:
Looks good.

Please, squash and add yourself to `AUTHORS`

jirutka:
Okay, done.

jesperpedersen:
Merged.

Could you up-port these changes to https://github.com/pgmoneta/pgmoneta and https://github.com/agroal/pgagroal as well ?

Issue #69, opened 2023-06-09
Query is supposed to run on both primary and replica servers


resyfer:
This [query](https://github.com/pgexporter/pgexporter/blob/main/contrib/yaml/postgresql-10.yaml#L14) is supposed to run on both `primary` and `replica` servers.

[Source](https://github.com/pgexporter/pgexporter/pull/67#discussion_r1217958888)

It can be removed entirely from the YAML file, as default value is `both`, however, since the whole file does not have any other examples with `server` in them, I'm tempted to mention it as `server: both`.

PR #70, opened 2023-06-09
[#69] Query's server value should be both

resyfer:


jesperpedersen:
Merged.

PR #73, opened 2023-07-01
[#72] Some empty queries break output

resyfer:


resyfer:
Rebased and force-pushed.

src/libpgexporter/prometheus.c
```diff
@@ -1246,15 +1232,16 @@ custom_metrics(int client_fd)
...
    while (temp)
    {
-      if (temp->query_alt->is_histogram)
-      {
-         handle_histogram(store, &n_store, temp);
-      }
-      else
-      {
-         handle_gauge_counter(store, &n_store, temp);
+      if (temp->error || (temp->query != NULL && temp->query->tuples != NULL)) {
```
jesperpedersen, line 1235: `temp->error` ?

`{` on wrong line - uncrustify

src/libpgexporter/prometheus.c
```diff
@@ -1421,6 +1407,10 @@ handle_histogram(column_store_t* store, int* n_store, query_list_t* temp)
       }
    }
 
+   if(!temp || !temp->query || !temp->query->tuples) {
```
jesperpedersen, line 1410: uncrustify

jesperpedersen (changes requested):

src/libpgexporter/prometheus.c
resyfer: Yeah I'll use uncrustify and push it, forgot about this sorry!

Also, `temp` was sometimes a result of an error, which caused the queries in it to remain empty. Since it is a singly linked list, it was a bit of a problem to free it and reuse it again, and the cause of one of the issues I'd mentioned. Any query that resulted in an error (unsupported, etc.) gets cleaned up later.

I simply took a note of the return value of executing the query on the server (1 if error, 0 if not) and put it `temp->error`.

resyfer:
I'm not sure why suddenly `pgexporter.h` was changed, even though I've used uncrustify every time I committed before 🤔

jesperpedersen:
Newer version ?

resyfer:
Oh yeah, I upgraded to fedora 38 by a clean install recently due to issues with 37.

```
╭─resyfer@fedora38 ~/pgexporter ‹null-query› 
╰─$ uncrustify --version
Uncrustify_d-0.76.0_f
```

jesperpedersen:
```
From https://github.com/resyfer/pgexporter
 * branch            null-query -> FETCH_HEAD
hint: Diverging branches can't be fast-forwarded, you need to either:
```

resyfer:
Does this help?

jesperpedersen:
Merged.

resyfer:
Thanks for the review!

PR #87, opened 2023-07-01
[#86] PostgreSQL V16 metrics examples

resyfer:


resyfer:
@jesperpedersen PTAL

contrib/yaml/postgresql-16.yaml
```diff
@@ -0,0 +1,199 @@
...
+        - name: confl_snapshot
+          type: counter
+          description: pg_stat_database_conflicts_confl_snapshot
+        - name: confl_bufferpin
+          type: counter
+          description: pg_stat_database_conflicts_confl_bufferpin
+        - name: confl_deadlock
+          type: counter
+          description: pg_stat_database_conflicts_confl_deadlock
+      version: 10
```
jesperpedersen, line 115: Better to do a comment section that lists all 10, like

```
#
# PostgreSQL 10
#
```

jesperpedersen (changes requested):

contrib/yaml/postgresql-16.yaml
resyfer: I didn't quite get this, can you please explain this more?

I've provided alternate versions of queries that are available in lower versions along with the alternative that is supported on the current version.

Also, should I make this change in every file?

jesperpedersen:
Rebase, please

resyfer:
@jesperpedersen

jesperpedersen:
Merged.

PR #89, opened 2023-07-02
[#88] More version specific Internal Metrics

resyfer:


jesperpedersen:
I'll look once the conflict is fixed, and pull request is rebased

resyfer:
@jesperpedersen I was keeping this PR for copying some examples from the various `postgresql-*.yaml` into `INTERNAL_YAML`.

I will look do some searching into what is preferred by people as necessary metrics out of the examples I've provided and then add those into `INTERNAL_YAML`.

Do you have any preference on the newly added metrics?

jesperpedersen:
I think this would be like the PostgreSQL 16 file , and pgexporter would use the correct part for each server

resyfer:
> I think this would be like the PostgreSQL 16 file , and pgexporter would use the correct part for each server

Yeah I'll take the metrics from there, but I don't think all of them should be taken? The output file does become quite big, and even though I don't think anyone will view the metrics output directly, I do have an inclination towards keeping it to as much as needed.

jesperpedersen:
Clear the draft status once you think it is ready

resyfer:
I've rectified my...very silly mistake....I didn't add the `\n` at the end of each string, forgetting that the strings will get concatenated one after the other 🥲 and I was wondering why it doesn't recognize the YAML syntax.

jesperpedersen:
Please, rebase

resyfer:
@jesperpedersen PTAL...by mistake I had rebased it with the grafana branch, but I removed the changes later which is why the deletion of files are shown.

jesperpedersen:
Merged.

Issue #90, opened 2023-07-20
Duplicate Server Name


resyfer:
Server names are allowed to be duplicates as of now.

eg.
```
[pgexporter]
host = *
metrics = 5002

log_type = file
log_level = info
log_path = /tmp/pgexporter.log

unix_socket_dir = /tmp/

[own]
host = localhost
port = 5432
user = pgexporter

[own]
host = localhost
port = 8010
user = pgexporter
```

PR #91, opened 2023-07-20
[#90] Duplicate Server Name

resyfer:
Warns the user about a name duplication (replication for config is provided in the issue):

image

It will still run as it did before, but the warning would hopefully encourage users to fix it.

### DIFF
--- a/src/libpgexporter/configuration.c
+++ b/src/libpgexporter/configuration.c
@@ -163,6 +163,17 @@ pgexporter_read_configuration(void* shm, char* filename)
                memcpy(&section, line + 1, max);
                if (strcmp(section, "pgexporter"))
                {
+
+                  for (int j = 0; j < idx_server - 1; j++)
+                  {
+                     if (!strcmp(srv.name, config->servers[j].name))
+                     {
+                        warnx("Duplicate server name \"%s\"", srv.name);
+                        fclose(file);
+                        exit(1);
+                     }
+                  }
+
                   if (idx_server > 0 && idx_server <= NUMBER_OF_SERVERS)
                   {
                      memcpy(&(config->servers[idx_server - 1]), &srv, sizeof(struct server));


resyfer:
@jesperpedersen PTAL and please let me know if I should let it be a warning or make it a fatal error and exit?

resyfer:
To use `pgexporter_log_error()` I'll have to make the following change (the change at the bottom is a bit iffy because of the return value if-statement involved).

image

So I can instead use the common way used in surrounding code.... ie. use `warnx()` to print the warning on screen, and then use `exit(1)`. I don't think logging is required to be started if the config is not correct, but I think that is just dependent on personal taste.

src/libpgexporter/configuration.c
```diff
@@ -163,6 +163,17 @@ pgexporter_read_configuration(void* shm, char* filename)
                memcpy(&section, line + 1, max);
                if (strcmp(section, "pgexporter"))
                {
+
+                  for (int j = 0; j < idx_server - 1; j++)
```
jesperpedersen, line 167: Why -1 ?

jesperpedersen (changes requested):

src/libpgexporter/configuration.c
resyfer: So, in `i`th iteration, the `i-1`th server (`srv`) is getting saved ([here](https://github.com/pgexporter/pgexporter/blob/21de355ec68e6da5e8ab6a86ef7ba0d4872a084a/src/libpgexporter/configuration.c#L187) and [here](https://github.com/pgexporter/pgexporter/blob/21de355ec68e6da5e8ab6a86ef7ba0d4872a084a/src/libpgexporter/configuration.c#L179)). So, if I need to compare `srv.name` to check duplication of names, I need to check for `0` to `i-2` including both.

resyfer:
@jesperpedersen Thank you for the review, but I need to make one more addition that just crossed my mind...the last `srv` is being copied outside the outermost `while` loop, so I need to make this addition as well to catch if the last server name provided in the config is a duplicate or not.

image

So sorry for overlooking this. Should I make a new PR for this? Or will you revert the merge?

jesperpedersen:
Just make a new one

Issue #92, opened 2023-07-21
Last server name in config can be duplicate


resyfer:
In the config, the last server provided as of now can have the same name as an existing previously defined server.

PR #93, opened 2023-07-21
[#92] Last server name in config can be duplicate

resyfer:


resyfer:
@jesperpedersen PTAL.

Additionally, even though `SERVER_UNDERTERMINED_VERSION` has a value of 0, and `memset` to 0 will give it that, I've relocated the line `SERVER_UNDERTERMINED_VERSION`

resyfer:
@jesperpedersen PTAL.

I've added the check for the last server name.

Also, I've moved the current name duplication checks inside the `if` statement, as if `idx_server` has a value more than `NUMBER_OF_SERVERS`, then it might lead to a segfault.

Also I moved the check of the last server to before the `memcpy` from my previous proposed position to just save some copy operations.

Additionally, even though `SERVER_UNDERTERMINED_VERSION` has a value of `0`, and `memset` to `0` will give it that, I've relocated the line with `SERVER_UNDERTERMINED_VERSION` to be near the location where `srv` members are initialized. This line might have remained from when I was cleaning the code up, and I was initially just thinking about removing it, but an explicit intialize of `srv.version` to `SERVER_UNDERTERMINED_VERSION` is easier to understand at a glance.

jesperpedersen:
Merged.

Issue #102, opened 2023-08-15
YAML Memory Leak


resyfer:
Memory leak about not clearing `event` always leading to direct leaks.

PR #103, opened 2023-08-15
[#102] YAML Memory Leak

resyfer:


jesperpedersen:
You need to rebase this...

jesperpedersen:
Please, explain the shared memory setup

resyfer:
> Please, explain the shared memory setup

So, the root of the respective AVL trees are at `config->prometheus[idx].root`.

The AVL trees are populated right at the start of the program, during YAML parsing, and remain in the memory till the very end, when they are cleared.

Due to the presence of child processes, these AVL trees will cause memory profilers to label them as memory leaks. Since the nodes need to be in shared memory to be used without giving `indirectly reachable` memory leak reports, each node has it's own memory allocated as a shared memory instead of malloc.

When it's time to "free" the memory of a node, the node has it's own allocated shared memory destroyed.

jesperpedersen:
Merged.

resyfer:
Thank you for the review!

Issue #115, opened 2023-10-18
`pgexporter-cli` command `is-alive` Invalid argument error


resyfer:
```
[pgexporter@fedora ~]$ pgexporter-cli -h 127.0.0.1 -p 8080 -U pgexporter -P pgexporter is-alive
2023-10-18 15:22:31 WARN  management.c:418 pgexporter_management_read_isalive: read: 3 Invalid argument
```

pgexporter Version 0.4.0
Platform: Fedora 38

PR #116, opened 2023-10-18
[#115] pgexporter-cli is-alive invalid argument

resyfer:


src/libpgexporter/management.c
```diff
@@ -422,6 +423,15 @@ pgexporter_management_read_isalive(SSL* ssl, int socket, int* status)
 
    *status = pgexporter_read_int32(&buf);
 
+   if (status)
```
resyfer, line 426: Should I make this `if(status == 1)` or is it fine?

jesperpedersen: We need to have the output on the `cli.c` side of things, and only if `-v` was passed

jesperpedersen (changes requested):

src/libpgexporter/management.c
resyfer: Hmmm, I was taking `pgexporter_management_read_status` as a reference. Also....what should be the output according to you if `-v` is not passed?

jesperpedersen: `management.c` shouldn't ideally have any output - the read functions should pass the values out to the caller... in some way.

For now, you can use `*status == 0`...

resyfer: > `management.c` shouldn't ideally have any output - the read functions should pass the values out to the caller... in some way.

Got it...so should I shift the output to [this place](https://github.com/pgexporter/pgexporter/blob/ff197819089d8a4bf7481a6f3075a6b7a86101b8/src/cli.c#L561)?

Also should I still go with printing the output for `-v` flag or should it be printed regardless?

> For now, you can use `*status == 0`...

Okay!

jesperpedersen: Yeah, ideally - in `cli.c` with `-v`.

They are in `management.c` because it was easier for the `..._read_` signatures.

We can do that in another pull request...

resyfer:
PTAL.

I also removed the `status != 2` part of the check as when pgexporter is running, it always puts in `1` in the response for status.

Also pgexporter is not executing, it gives off an error like:

```
pgexporter-cli: pgexporter-cli: No route to host: localhost:8080
munmap_chunk(): invalid pointer
Aborted (core dumped)
```

which may be an issue that needs be fixed as well.

jesperpedersen:
Yeah, that is something that needs to be fixed

resyfer:
> Yeah, that is something that needs to be fixed

Sure, I'll look into this. Do you want me to continue in this PR or a new PR & issue?

jesperpedersen:
Just continue in this one

resyfer:
I did a very basic patch which was basically that if the user provided password in the CLI arguments, and if `pgexporter` wasn't running, it would bypass the `do_free = false` part.

There's still an issue that they way the current code is, it will output:

```
pgexporter-cli: pgexporter-cli: No route to host: localhost:8080
```

I can add an if-statement inside this warning for this specific case, but I wonder if it will become unintuitive in the code...and any rewrite of the code will require a LOT of changes.

jesperpedersen:
So... what are you thinking about this ?

resyfer:
@jesperpedersen I had to sort of create an exception (the special-case type) in the error handling, as in, basically allowing checking if pgexporter is alive, and it actually not being alive, is not an error, but rather a special case.

I've thought about this for some time, and while this solution is a simple duct-tape like patch, everything else I've thought of are unnecessarily complex, and this keeps the code fairly readable as well.

src/libpgexporter/management.c
```diff
@@ -394,8 +395,10 @@ pgexporter_management_isalive(SSL* ssl, int socket)
 {
    if (write_header(ssl, socket, MANAGEMENT_ISALIVE))
    {
-      pgexporter_log_warn("pgexporter_management_isalive: write: %d", socket);
-      errno = 0;
+      // No error or warning for checking if alive
+
+      // pgexporter_log_warn("pgexporter_management_isalive: write: %d", socket);
```
jesperpedersen, line 400: You can remove these lines

src/cli.c
```diff
@@ -453,7 +468,10 @@ main(int argc, char** argv)
 
    if (do_free)
    {
-      free(password);
+      if (password)
```
jesperpedersen, line 471: You don't need to guard a `free()` - it understands a `NULL` value

jesperpedersen (changes requested):

src/cli.c
resyfer: Got it.

jesperpedersen:
Rebased, merged and back-ported.

Issue #117, opened 2023-10-18
`pgexporter-cli` memory bug on multiple `status`/`details`


resyfer:
Terminal 1:
```
[pgexporter@fedora ~]$ pgexporter -Y ./y/postgresql-10.yaml -c ./pgexporter.conf -u ./pgexporter_users.conf -A ./pgexporter_users.conf
free(): double free detected in tcache 2
free(): invalid pointer
free(): double free detected in tcache 2
Aborted (core dumped)
```

Terminal 2:
```
[pgexporter@fedora ~]$ pgexporter-cli details -h 127.0.0.1 -p 8080 -U pgexporter -P pgexporter
Number of servers: 8
Server           : own
  Active         : Yes
Server           : v10
  Active         : No
Server           : v11
  Active         : No
Server           : v12
  Active         : No
Server           : v13
  Active         : No
Server           : v14
  Active         : No
Server           : v15
  Active         : No
Server           : v16beta2
  Active         : No
[pgexporter@fedora ~]$ pgexporter-cli details -h 127.0.0.1 -p 8080 -U pgexporter -P pgexporter
Number of servers: 8
Server           : own
  Active         : Yes
Server           : v10
  Active         : No
Server           : v11
  Active         : No
Server           : v12
  Active         : No
Server           : v13
  Active         : No
Server           : v14
  Active         : No
Server           : v15
  Active         : No
Server           : v16beta2
  Active         : No
[pgexporter@fedora ~]$ pgexporter-cli details -h 127.0.0.1 -p 8080 -U pgexporter -P pgexporter
2023-10-18 21:54:55 WARN  management.c:498 pgexporter_management_read_status: read: 3 Invalid argument
```

Version: 0.4.0
Fedora 38

PR #118, opened 2023-10-18
[#117] pgexporter-cli memory bug on multiple status/details

resyfer:


resyfer:
I'll rebase this after you merge #115. PTAL.

src/libpgexporter/management.c
```diff
@@ -319,7 +319,6 @@ pgexporter_management_write_status(int socket)
 {
    configuration_t* config;
 
-   pgexporter_start_logging();
```
resyfer, line 321: I removed these as in all situations (in main), logging was started before this function is run, and sometime after this function is finished.

Starting and closing logging inside creates the problems:

1. fclose(log_file) had no check for `log_file == NULL` (I added this).
2. `pgexporter_stop_logging` is called in main again later, which leads to double free (or closing of logging prematurely, just because the user made a `status`/`details` request using `pgexporter-cli`).

Because of these, I removed logging from these functions.

jesperpedersen: We need to investigate how we can do logging in `cli.c` as well...

jesperpedersen:
So, what are you thinking about it ?

resyfer:
@jesperpedersen PTAL

jesperpedersen:
Merged.

Issue #119, opened 2023-10-27
Quotes in `query` label of `pgexporter_pg_stat_statements_*` not escaped


cracksalad:
**Describe the bug**

In the metrics output I find:
```
pgexporter_pg_stat_statements_calls{server="my_database",query="INSERT INTO "measurement" ("sensor_uuid", "value", "timestamp", "tags") VALUES($1,$2,$3,$4)"} 35400
```
As you can see, there are double quotes in the value of the `query` label which are not escaped and therefore break the syntax.

As far as I can tell, the following metrics are affected:

- `pgexporter_pg_stat_statements_calls`
- `pgexporter_pg_stat_statements_rows`
- `pgexporter_pg_stat_statements_total_exec_time`
- `pgexporter_pg_stat_statements_plans`
- `pgexporter_pg_stat_statements_wal_bytes`
 
... basically every pg_stat_statements metric with query label.

**To Reproduce**

1. Enable pg_stat_statements extension in your database
2. Use double quotes to escape table and column names in your queries

**Version**

Tested with release 0.4.0 as well as branch 0.4.x

**PostgreSQL**

14

**libev**

4.33

**OpenSSL**

3.0.7

**OS**

AlmaLinux 9.2

**Configuration**

pgexporter.conf:

```ini
[pgexporter]
host = 10.8.1.4
metrics = 5002

log_type = file
log_level = info
log_path = /var/log/pgexporter.log

unix_socket_dir = /tmp/

[my_database]
host = 127.0.0.1
port = 5432
user = pgexporter
```

**Prometheus**

Can you provide the output of the invalid Prometheus metrics ?

See above. There are of course multiple examples. If you need more to reproduce or understand the bug, just ask.

jesperpedersen:
@resyfer Can you take a look at this ?

cracksalad:
Could we just replace `SELECT calls, query` by `SELECT calls, REGEXP_REPLACE(query, '"', '\"') AS query` in [contrib/yaml/postgresql-XX.yaml](https://github.com/pgexporter/pgexporter/blob/775cb52ee815882695f7ab728922e7a376ca16db/contrib/yaml/postgresql-14.yaml#L506)?

jesperpedersen:
Did you try that locally ?

cracksalad:
> Did you try that locally ?

[...]

PR #120, opened 2023-11-22
Draft double quote escaping for #119

cracksalad:
See #119 

@jesperpedersen @resyfer 

This would be a working solution. WDYT?

I think there are two left issues:

1. My double escaping avoidance does not do the trick. Think about the string `\\"`. If only checking the character in front of `"`, it would think the double quotes are already escaped, but they are not. What if we escape not only double quotes but also backslashes? 🤔 
2. Since a new memory allocation is done even if no escaping is necessary, performance is probably not perfect. But how to know if one has to free, because it was allocated or not, because escaping was not necessary?


resyfer:
> This would be a working solution. WDYT?

LGTM with some small changes that I feel need to be there:

#### #1
[Here](https://github.com/pgexporter/pgexporter/pull/120/files#diff-205f40cdedd72fcb0fd96b2d26630c05e96c6eb2f684e5f5881b3e9d8a616e91R1969), I feel the integer returned from `safe_prometheus_key_additional_length(key)` should be calculated first, and only on being non-zero, should new memory be allocated, something like this.
 
Since you're already modifying every place safe_prometheus_key is being used, I do _feel_ you should instead change it to this (@jesperpedersen need your approval on this):

```c
/*
 * @returns 0 if all right, 1 otherwise.
 */
static int
safe_prometheus_key(char** key) {
   int extra_len = safe_prometheus_key_additional_length(key);

   if (extra_len) {
      return 0;
   }

   char escaped_key = /* malloc */;

   // Escaping stuff

   safe_prometheus_key_free(*key); // Free up old stuff
   *key = escaped_key;

   return 0;
}
```

Keys are, as of my knowledge, always dynamically allocated, so you can just free them and replace it with new stuff. 

#### #2 
I feel the name should be a bit more clear than `safe_key1` and `safe_key2` [here](https://github.com/pgexporter/pgexporter/pull/120/files#diff-205f40cdedd72fcb0fd96b2d26630c05e96c6eb2f684e5f5881b3e9d8a616e91R1057).

---

> 1. My double escaping avoidance does not do the trick. Think about the string `\\"`. If only checking the character in front of `"`, it would think the double quotes are already escaped, but they are not. What if we escape not only double quotes but also backslashes? 🤔

Can you give an example for this? I did not quite get what you mean.

From what I understood from your sentence, and from what I understand of Prometheus' keys, backslashes `\` within text (ie. within a pair of `"`) do not need to be escaped (the [docs](https://prometheus.io/docs/instrumenting/writing_exporters/#labels) are really not too clear about it afaik).

jesperpedersen:
Remember to update `AUTHORS` as well

cracksalad:
@resyfer Thank you for your comprehensive feedback!

> Since you're already modifying every place safe_prometheus_key is being used, I do feel you should instead change it to this (@jesperpedersen need your approval on this)

Then I am awaiting approval on this.

> Keys are, as of my knowledge, always dynamically allocated (you can see [here](https://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/queries.c#L540) and [here](https://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/queries.c#L783)), so you can just free them and replace it with new stuff.

From what I have seen so far, the key is indeed always dynamically allocated. However I am hesitating to free the key because it might be referenced elsewhere. Are you sure about that?
If we already change the parameter type to `char**`, we should be able to use `realloc` which is hopefully safer. However I am not very experienced with this.

> I feel the name should be a bit more clear than safe_key1 and safe_key2 [here](https://github.com/pgexporter/pgexporter/pull/120/files#diff-205f40cdedd72fcb0fd96b2d26630c05e96c6eb2f684e5f5881b3e9d8a616e91R1057).

Is escaping for the value of `safe_key2` needed at all? It seems to be a help text inside a comment. Escaping '#' might be useful, but that is not implemented. Escaping dots and double quotes does not seem to make any sense.
If I am wrong, what would you suggest as a name?

> Can you give an example for this? I did not quite get what you mean.

Sure! Lets say your label value (string's content) is `hello \"world\"!` already before calling `safe_prometheus_key`. Right now this would be converted to `hello \\"world\\"!` which contains unescaped double quotes again. Even if this looks like an edge case, if you realize the escaping inside your queries, those cases are real. 
One solution would be to count the directly preceding backslashes in front of each double quotes and if the number is even... too complicated!
Another solution might be to escape backslashes as well, which leads to `hello \\\"world\\\"!` in this example. This is probably not what you wanted, when you realized the escaping inside the SQL queries, but at least it does not break the Prometheus syntax.

Feel free to throw your opinion on all of this.

resyfer:
> Then I am awaiting approval on this.

@jesperpedersen 

> we should be able to use `realloc` which is hopefully safer. However I am not very experienced with this.

Hmmm, both are equally fine I should think...it's just that realloc just requires mental gymnastics for escaping...but both are fine.

---

> Is escaping for the value of `safe_key2` needed at all? It seems to be a help text inside a comment. Escaping '#' might be useful, but that is not implemented.

The Prometheus docs are not really clear on this, but the text that you're using with `safe_key1` & `safe_key2` is basically:

```
#HELP pgexporter_metric_name_safe_key1 safe_key2
#TYPE pgexporter_metric_name_safe_key1 gauge
```

These provide the help text and type info of a metric to prometheus.

It's not clear if `safe_key2` is required to escape. `safe_key1` is fine even with the previous implementation. I think it's fine the way it is.

---

Also I think you should uncomment the part where you check if a `"` is already escaped. Making the condition `(key[i] == '"' && (i == 0 || (i > 0 && k[i-1] != '\\')))` should be fine I think.

---

> Sure! Lets say your label value (string's content) is `hello \"world\"!` already before calling `safe_prometheus_key`. Right now this would be converted to `hello \\"world\\"!` which contains unescaped double quotes again. Even if this looks like an edge case, if you realize the escaping inside your queries, those cases are real. One solution would be to count the directly preceding backslashes in front of each double quotes and if the number is even... too complicated! Another solution might be to escape backslashes as well, which leads to `hello \\\"world\\\"!` in this example. This is probably not what you wanted, when you realized the escaping inside the SQL queries, but at least it does not break the Prometheus syntax.

Got it, thanks for the explanation.

I think the counter with the same condition (as I mentioned above) will suffice:

```c
if (key[i] == '"' && (i == 0 || (i > 0 && k[i-1] != '\\'))) {
   escape_counter++;
}
```

cracksalad:
Just added the escaping of backslashes as well, since it is more stable than just checking the char before double quotes and it should not bring any downsides. 

___

Additionally I removed the escaping of the help texts, since it is not necessary (right?) and with escaping some help texts looked like this:
```
#HELP pgexporter_pg_settings_backslash_quote Sets whether \"\\'\" is allowed in string literals.
#TYPE pgexporter_pg_settings_backslash_quote gauge
```
which I think is not that readable. 

> It's not clear if safe_key2 is required to escape.

[safe_key2 is used as part of the help text]

At least double quotes do not have to be escaped in help texts as the example above shows, after disabling the escaping.

___

If @jesperpedersen approves [this proposed change](https://github.com/pgexporter/pgexporter/pull/120#issuecomment-1823293982) I will implement it. Otherwise this PR is now complete, from my point of view.

___

Please let me know what you think!

jesperpedersen:
@cracksalad I think it is ok to go forward with this - we can always expand escaping in additional patches. Can you squash your commits ?

cracksalad:
@jesperpedersen successfully squashed! First time I did this manually 😄 

I have seen that it is also possible to configure the repo to squash commits of PRs automatically. This might be easier for the future. See [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests)

src/libpgexporter/prometheus.c
```diff
@@ -1909,32 +1933,74 @@ get_value(char* tag, char* name, char* val)
    return "1";
 }
 
+static int
+safe_prometheus_key_additional_length(char* key)
+{
+   int count = 0;
+   int i = 0;
+   
+   while(key[i] != '\0')
```
jesperpedersen, line 1942: uncrustify.sh

src/libpgexporter/prometheus.c
```diff
@@ -1909,32 +1933,74 @@ get_value(char* tag, char* name, char* val)
...
+   {
+      if(key[i] == '"' || key[i] == '\\')
```
jesperpedersen, line 1944: uncrustify.sh

src/libpgexporter/prometheus.c
```diff
@@ -1909,32 +1933,74 @@ get_value(char* tag, char* name, char* val)
...
    int i = 0;
+   int j = 0;
 
-   if (key == NULL)
+   if (key == NULL || strlen(key) == 0)
    {
       return "";
    }
 
+   char* escaped = (char*) malloc(strlen(key) + safe_prometheus_key_additional_length(key) + 1);
```
jesperpedersen, line 1966: Move declaration of `escaped` to beginning of function, and only assign here

src/libpgexporter/prometheus.c
```diff
@@ -1909,32 +1933,74 @@ get_value(char* tag, char* name, char* val)
...
+      j++;
+   }
+   escaped[j] = '\0';
+   return escaped;
+}
+
+static void 
+safe_prometheus_key_free(char* key)
+{
+   if (strlen(key) > 0)
```
jesperpedersen, line 2000: Missing `NULL` check

jesperpedersen (changes requested):

jesperpedersen:
>I have seen that it is also possible to configure the repo to squash commits of PRs automatically. This might be easier for the future. See [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests)

Yeah, but we need Squash, Rebase, Merge - so all 3, and GitHub doesn't support that.

PR #134, opened 2024-04-25
Fix prompts not being printed by disabling stdout buffering

jirutka:
When built against musl libc, prompts like "Master key:", "Username:" etc. are not printed to console because of stdout buffering.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -118,6 +118,9 @@ main(int argc, char** argv)
    int option_index = 0;
    int32_t action = ACTION_UNKNOWN;
 
+   // Disable stdout buffering (i.e. write to stdout immediatelly).
+   setbuf(stdout, NULL);
+
    while (1)
    {
       static struct option long_options[] =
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Would you be interested in porting over the improvements that was done in pgmoneta to pgexporter 

https://github.com/pgmoneta/pgmoneta/blob/main/src/cli.c
https://github.com/pgmoneta/pgmoneta/blob/main/src/admin.c

?

Or going with your current patch...

jirutka:
I don’t understand, which changes from pgmoneta to port and how is it related to this issue?

jesperpedersen:
The same issue should be fixed in pgmoneta...

But, we can go with your patch too

resyfer (approved):

resyfer:
TL;DR LGTM.

A `NULL` value is same as `_IONBF` acc to [man page](https://man7.org/linux/man-pages/man3/setbuf.3.html), so all good. `pgexporter` does not print too much to `stdout` for this to impact performance imo, so this should be fine instead of `fflush`.

jesperpedersen:
Merged.

PR #196, opened 2025-04-01
[#175] feat: implement prometheus support using art

### LINKED ISSUE #175: Implement prometheus support using ART
**Is your feature request related to a problem? Please describe.**

Just like `prometheus_bridge.c` works using ART, allowing timestamps and sorting, can `prometheus.c` use ART as well? Will there be any issues or performance hits?

botirk38:
Fixes #175


botirk38:
> Why did you use `prometheus_bridge` ? It should be ART per category

I basically copied bridge one for one, I thought that was the correct strategy but fit it within the context of the prometheus file, could your clarify further?

botirk38:
> Why did you use `prometheus_bridge` ? It should be ART per category

OH I think I get it you mean there should be a separate art for each metric instead of just one for all is that it?

botirk38:
> Yes, create each "block" of metrics in their own ART

```c
struct prometheus_categories {
   struct art** categories;  // Dynamic array of ARTs
   int count;                // Number of categories
};
```

something like this?

jesperpedersen:
Just start with an ART within each category - or start a discussion on the best design on it

botirk38:
> Just start with an ART within each category - or start a discussion on the best design on it

Please review my code again, based on the changes we discussed

botirk38:
> Create `static` functions to help you with creating a `prometheus_metric` f.ex.
> 
> We don't need the timestamp for this code so it is just `= 0`.
> 
> And, use `pgexporter_append()` instead of `strdup()`

Ok couple of questions:

Why do we not need timestamps so that I can understand?
Is the general approach with the art per metric category ok with you, the idea was I write some quick code see if the general approach is ok before I clean it up, to waste less time

jesperpedersen:
There is no timestamp in the output, so we only need to update the timestamp for the cache (if enabled).

Yes, doing an ART inside each category is good

src/libpgexporter/prometheus.c
```diff
@@ -693,949 +683,1153 @@ collector_pass(const char* collector)
...
+            memset(attr, 0, sizeof(struct prometheus_attribute));
+
+            attr->key = strdup(temp->query_alt->columns[j].name);
+
+            /* Use safe_prometheus_key for label values */
+            char* safe_key = safe_prometheus_key(pgexporter_get_column(j, tuple));
+            attr->value = strdup(safe_key);
+            safe_prometheus_key_free(safe_key);
+
+            pgexporter_deque_add(attrs->attributes, NULL, (uintptr_t)attr, ValueRef);
```
Jubilee101, line 1793: You are adding attr as ValueRef, so deque won't destroy any of it. I've only taken a glance so I'm not sure if this is on purpose. If not you are going to have mem leak. Have you run clang with debug mode to verify?

botirk38: I ran CMake with the following command:

```
cmake -DCMAKE_BUILD_TYPE=Debug ..
```

I found no issues in the output.

jesperpedersen:
@Jubilee101 PTAL

botirk38:
> @Jubilee101 PTAL

Any updates?

jesperpedersen:
@botirk38 Remember to rebase, squash, and force push

src/libpgexporter/prometheus.c
```diff
@@ -2551,3 +2687,305 @@ metrics_cache_finalize(void)
...
+      send_chunk(ssl,client_fd, data);
+      free(data);
+   }
+
+   pgexporter_art_iterator_destroy(metrics_iterator);
+}
+
+/* Common function to clean up an ART */
+static void
+cleanup_metrics_art(struct art* metrics_art)
```
Jubilee101, line 2785: Have you checked out `pgexporter_art_insert_with_config`? You can register a destroy callback for `metric` or `attribute`, then you don't have go to through all this trouble of destroying a deeply nested ART.

jesperpedersen:
@botirk38 Yes ?

botirk38:
> @botirk38 Remember to rebase, squash, and force push

Oh yes sorry. That will take some time to do because I had to merge from main.

src/libpgexporter/prometheus.c
botirk38: yea sure makes sense will look into it.

botirk38:
> @botirk38 Still working on this ?

yes, give me till next week, wrapping up some stuff.

PR #209, opened 2025-04-08
[#204] Add version coverage for clang and gcc

bassamadnan:
Added the sanitizer flags for Debug builds with version checking, new flags added are 
1. `-fsanitize-address-use-after-scope`: helps detecting when a variable is accessed after it has gone out of scope such as when a pointer to a local variable continues to be used after the function returns or when a reference to a variable in an inner scope is accessed after that scope ends.
2. `-fsanitize-recover=address`: continues execution after error detection.
3. `-Wextra` for extra warnings around operators
Rest are unchanged, heres a minimum version report for each of the flags.

<details>
<summary>Clang Sanitizer Flags</summary>

| Flag | Minimum Version |
|------|-----------------|
| -fsanitize=address | 3.2 |
| -fno-sanitize=null | 3.2 |
| -fsanitize=undefined | 3.3 |
| -fsanitize=float-divide-by-zero | 3.3 |
| -fsanitize=float-cast-overflow | 3.3 |
| -fno-sanitize-recover=all | 3.6 |
| -fsanitize-recover=address | 3.8 |
| -fsanitize-address-use-after-scope | 3.9 |
| -fno-sanitize=alignment | 3.2 |
</details>

<details>
<summary>GCC Sanitizer Flags</summary>

| Flag | Minimum Version |
|------|-----------------|
| -fsanitize=address | 4.8 |
| -fsanitize=undefined | 4.9 |
| -fno-sanitize=alignment | 5.1 |
| -fno-sanitize-recover=all | 5.1 |
| -fsanitize=float-cast-overflow | 5.1 |
| -fsanitize=float-divide-by-zero | 5.1 |
| -fsanitize-recover=address | 6.0 |
| -fsanitize-address-use-after-scope | 7.0 |
</details>

These were added by testing them on [godbolt](https://godbolt.org/). They provide their APIs so I sent it requests for each of the versions available. Apart from that I manually downloaded the source code from LLVM repositry (for their flags and sanitzer args) and from the gcc repositry and wrote a script to check these flags using string matching accross different version files. These can be found [here](https://github.com/bassamadnan/asan-clang). I have cross checked the results from godbolt as well.

The implementation of clang is abit different from gcc because clang depeends on the order in which these flags are added otherwise we run into linking issues, and the version check is done for the minimum version after which it has continously been supported (some flags in clang have disruptions where they throw errors). This wouldn't matter since the compiler version which has support for C17 will not run into these issues. I hope this looks fine.


jesperpedersen:
@Jubilee101 Cc:

jesperpedersen:
@bassamadnan Can you look at the warnings ? Some should be fixed, and some should be ignored

bassamadnan:
@jesperpedersen do you mean the Wall/Wextra warnings or are there some make warnings?

bassamadnan:
@jesperpedersen yes i would be happy to take a look at them! I think priority for now is
1. After this PR is reviewed port it to pgmoneta and pgagroal
2. Work on fixing the warnings side by side currently

I am using Clang 20+  and i do see quite a few errors; i think we can have an issue for it? I will ask to be assigned. Will work on them accross the next few days and crosscheck with gcc too whenever i can. I hope this sounds fine

jesperpedersen:
Just do them as part of this

bassamadnan:
@jesperpedersen There's alot of comparision related errors, should i typecast it or redefine the datatypes? also a few unused variable errors, should i just remove them?

jesperpedersen:
Redefine.

The unused are likely due to APIs where they are necessary - so you need to add an ignore if that is the case

bassamadnan:
`__off_t` isn't defined for mac, will revert it

bassamadnan:
I added unused attribute so compiler excludes the parameters from warnings. [Some](https://github.com/pgexporter/pgexporter/blob/4110d0a3d579128f33cd24614d9bc91ed34ce171/src/libpgexporter/connection.c#L280) places are hard to fix without refactoring, if I naively typecast `size` variable to ssize_t we may fall into range issues, and typecasting `numbytes` variable to size_t isn't trivial without refactoring. I did typecast at one other place where it was safe to do so and redefined everywhere else.

src/libpgexporter/security.c
```diff
@@ -2752,7 +2752,7 @@ pgexporter_extract_server_parameters(struct deque** server_parameters)
          data = &security_messages[i][0];
          offset = 0;
 
-         while (offset < data_length)
+         while (offset < (size_t) data_length)
```
bassamadnan, line 2755: This is the only place i've typcasted at

bassamadnan: Hi, i fixed whatever looked fine. Could you please have a look ?  @jesperpedersen .

jesperpedersen:
I'm getting warns from -Wsign-compare

jesperpedersen:
gcc says

```
[ 86%] Linking C shared library libpgexporter.so
/usr/bin/ld: cannot find /usr/lib64/libasan.so.8.0.0: No such file or directory
collect2: error: ld returned 1 exit status
```

jesperpedersen:
```
[ 86%] Linking C shared library libpgexporter.so
/usr/bin/ld: cannot find /usr/lib64/libasan.so.8.0.0: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [src/CMakeFiles/pgexporter.dir/build.make:588: src/libpgexporter.so.0.7.0] Error 1
make[1]: *** [CMakeFiles/Makefile2:259: src/CMakeFiles/pgexporter.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

jesperpedersen:
> edit: for gcc alot of them come from our macro usage of MIN, should be safe to ignore this i feel
Then we need to disable that check because we can't have warns in the output, so see if you can fix them

jesperpedersen:
Ah, it is 2 packages extra

```
libasan libasan-static
```

That need to be documented

jesperpedersen:
Yeah, just remove sign-compare for both, and unused-parameter for gcc

bassamadnan:
> Ah, it is 2 packages extra
> 
> ```
> libasan libasan-static
> ```
> 
> That need to be documented

I'll add that in this PR, in the same .md i have added.  I don't explicitly remember installing it; they mightve come with the compiler. Glad this worked though i was abit freaked out.

src/libpgexporter/message.c
```diff
@@ -827,6 +827,7 @@ ssl_read_message(SSL* ssl, int timeout, struct message** msg)
                   /* Sleep for 100ms */
                   SLEEP(100000000L);
                }
+               __attribute__((fallthrough));
```
bassamadnan, line 830: put this here to remove the fall through warning, i think a `break;` is supposed to be here

bassamadnan: I kept the unused parameter warning for both. I realized the way gcc/clang handles __attribute is abit different so I just kept it in a standard way. No more errors now.

jesperpedersen:
Do a `git grep "dnf install"` and add the libasan packages there...

It seems that a lot need to be sync'ed with README.md (and PostgreSQL 17) too...

bassamadnan:
is this fine? @jesperpedersen

CONTRIBUTING.md
```diff
@@ -37,7 +37,7 @@ Follow [this](./doc/manual/user15-docker.md) to build using docker
 For Red Hat RPM based distributions use the following command:
 
 ```
-dnf install git gcc cmake make libev libev-devel openssl openssl-devel systemd systemd-devel python3-docutils
+dnf install git gcc cmake make libev libev-devel openssl openssl-devel systemd systemd-devel python3-docutils libasan libasan-static
```
jesperpedersen, line 40: Missing some dependencies - just copy from README.md

doc/manual/dev-03-rpm.md
```diff
@@ -7,7 +7,7 @@
 ## Requirements
 
 ```sh
-dnf install gcc rpm-build rpm-devel rpmlint make python bash coreutils diffutils patch rpmdevtools chrpath
+dnf install gcc rpm-build rpm-devel rpmlint make python bash coreutils diffutils patch rpmdevtools chrpath libasan libasan-static
```
jesperpedersen, line 10: We don't need it here

src/libpgexporter/logging.c
```diff
@@ -485,7 +485,7 @@ pgexporter_log_mem(void* data, size_t size)
                j++;
                k = 0;
             }
-            if (c >= 32 && c <= 127)
+            if (c >= 32)
```
jesperpedersen, line 488: Why ?

src/libpgexporter/message.c
```diff
@@ -82,7 +82,7 @@ pgexporter_write_message(SSL* ssl, int socket, struct message* msg)
 }
 
 void
-pgexporter_clear_message(struct message* msg)
+pgexporter_clear_message(struct message* msg __attribute__((unused)))
```
jesperpedersen, line 85: You can do `(void)`

jesperpedersen (changes requested):

src/libpgexporter/message.c
bassamadnan: okay i'll reflect that everywhere, avoided this since it would add 2-3  lines to alot of functions

jesperpedersen: Just for this one

src/libpgexporter/logging.c
jesperpedersen: Ok

src/libpgexporter/message.c
```diff
@@ -84,6 +84,7 @@ pgexporter_write_message(SSL* ssl, int socket, struct message* msg)
 void
 pgexporter_clear_message(struct message* msg)
 {
+   (void) msg;
```
jesperpedersen, line 87: I meant `pgmoneta_clear_message(void)`

jesperpedersen (changes requested):

src/libpgexporter/message.c
bassamadnan: oh okay

src/CMakeLists.txt
```diff
@@ -170,16 +170,75 @@ if (CMAKE_BUILD_TYPE MATCHES Debug)
...
+            add_compile_options(-fsanitize-recover=address)
+          endif()
+          
+          if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.9)
+            add_compile_options(-fsanitize-address-use-after-scope)
+          endif()
+          
+          add_compile_options(-fno-sanitize=alignment)
+
+          set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS_DEBUG} -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow -fno-sanitize=null -fno-sanitize=alignment")
```
jesperpedersen, line 197: You can use https://cmake.org/cmake/help/v3.14/command/add_link_options.html too in each section

src/include/message.h
```diff
@@ -90,7 +90,7 @@ pgexporter_write_message(SSL* ssl, int socket, struct message* msg);
  * @param msg The resulting message
```
jesperpedersen, line 90: Documentation

jesperpedersen (changes requested):

jesperpedersen:
@bassamadnan Otherwise it looks ok

src/CMakeLists.txt
bassamadnan: tried it, among other things, it just does not work and fails to find refferences (linking).

<details>
  <summary>Code block for Debug section</summary>

  ```
  # Debug
  if (CMAKE_BUILD_TYPE MATCHES Debug)
    add_compile_options(-O0)
    add_compile_options(-DDEBUG)
  
    if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
      if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
        if (NOT ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
          if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.2)
            add_compile_options(-fsanitize=address)
            add_compile_options(-fno-sanitize=null)
            add_link_options(-fsanitize=address -fno-sanitize=null)
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.3)
              add_compile_options(-fsanitize=undefined)
              add_compile_options(-fsanitize=float-divide-by-zero)
              add_compile_options(-fsanitize=float-cast-overflow)
              add_link_options(-fsanitize=undefined -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow)
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.6)
              add_compile_options(-fno-sanitize-recover=all)
              add_link_options(-fno-sanitize-recover=all)
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.8)
              add_compile_options(-fsanitize-recover=address)
              add_link_options(-fsanitize-recover=address)
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 3.9)
              add_compile_options(-fsanitize-address-use-after-scope)
              add_link_options(-fsanitize-address-use-after-scope)
            endif()
            
            add_compile_options(-fno-sanitize=alignment)
            add_link_options(-fno-sanitize=alignment)
          endif()
        endif()
      endif()
    elseif(CMAKE_C_COMPILER_ID STREQUAL "GNU")
      if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
        if (NOT ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
          if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 4.8)
            add_compile_options(-fsanitize=address)
            
            set(TEMP_EXE_LINKER_FLAGS "-fsanitize=address")
            set(TEMP_SHARED_LINKER_FLAGS "-fsanitize=address")
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 4.9)
              add_compile_options(-fsanitize=undefined)
              set(TEMP_EXE_LINKER_FLAGS "${TEMP_EXE_LINKER_FLAGS} -fsanitize=undefined")
              set(TEMP_SHARED_LINKER_FLAGS "${TEMP_SHARED_LINKER_FLAGS} -fsanitize=undefined")
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 5.1)
              add_compile_options(-fno-sanitize=alignment)
              add_compile_options(-fno-sanitize-recover=all)
              add_compile_options(-fsanitize=float-cast-overflow)
              add_compile_options(-fsanitize=float-divide-by-zero)
              
              set(TEMP_EXE_LINKER_FLAGS "${TEMP_EXE_LINKER_FLAGS} -fno-sanitize=alignment -fno-sanitize-recover=all -fsanitize=float-cast-overflow -fsanitize=float-divide-by-zero")
              set(TEMP_SHARED_LINKER_FLAGS "${TEMP_SHARED_LINKER_FLAGS} -fno-sanitize=alignment -fno-sanitize-recover=all -fsanitize=float-cast-overflow -fsanitize=float-divide-by-zero")
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 6.0)
              add_compile_options(-fsanitize-recover=address)
              set(TEMP_EXE_LINKER_FLAGS "${TEMP_EXE_LINKER_FLAGS} -fsanitize-recover=address")
              set(TEMP_SHARED_LINKER_FLAGS "${TEMP_SHARED_LINKER_FLAGS} -fsanitize-recover=address")
            endif()
            
            if (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
              add_compile_options(-fsanitize-address-use-after-scope)
              set(TEMP_EXE_LINKER_FLAGS "${TEMP_EXE_LINKER_FLAGS} -fsanitize-address-use-after-scope")
              set(TEMP_SHARED_LINKER_FLAGS "${TEMP_SHARED_LINKER_FLAGS} -fsanitize-address-use-after-scope")
            endif()
            
            set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${TEMP_EXE_LINKER_FLAGS}")
            set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${TEMP_SHARED_LINKER_FLAGS}")
          endif()
        endif()
      endif()
    endif()
  
    check_c_compiler_flag(-fno-omit-frame-pointer HAS_NO_OMIT_FRAME_POINTER)
    if (HAS_NO_OMIT_FRAME_POINTER)
      set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-omit-frame-pointer")
    endif()
    
    check_c_compiler_flag(-Wextra HAS_WEXTRA)
    if (HAS_WEXTRA)
      add_compile_options(-Wextra)
      
      check_c_compiler_flag(-Wno-sign-compare HAS_NO_SIGN_COMPARE)
      if (HAS_NO_SIGN_COMPARE)
        add_compile_options(-Wno-sign-compare)
      endif()
    endif()
  endif()
  ```
</details>
I've tried putting it in a list and setting flags to that list as well (similliar to gcc) that fails as well. I've tried to link  them specifically at `target_link_libraries` towards the end of the file, that fails too.  It should be harmless though since the linker flags will likely just be ignored, especially for clang versions witch C17 support all flags are just present often.

src/include/message.h
bassamadnan: Apologies, had overlooked this.

jesperpedersen:
Merged.

jesperpedersen:
@bassamadnan Please, port this patch to pgagroal and pgmoneta

bassamadnan:
>  port this patch to pgagroal and pgmoneta
yessir will be on it, i'll fix the warnings produced too (first without disabling the sign compare flag) wherever I can.

Issue #213, opened 2025-04-26
Scraper script for a Prometheus endpoint


jesperpedersen:
Create a bash script that will extract the Prometheus and their documentation including additional information for each metric

It could be

* Table of Content
* Metric name
* Metric description
* Additional information
* An example

jesperpedersen:
[prometheus.txt](https://github.com/user-attachments/files/19923737/prometheus.txt) could be used as a starting point

bassamadnan:
We could have the value we fetch from the endpoint as the example right? Can I take this up? I'll focus on the core metrics.

I'd also like to understand the usecase abit- what would this be used for?

jesperpedersen:
It is to keep documentation in sync with the actual endpoint without having to manually check everything. Try the PoC script against an installation

bassamadnan:
yes, heres the current MD generated
<details>
<summary>PoC setup</summary>

### pgexporter_logging_fatal

The number of FATAL logging statements

### pgexporter_version

The pgexporter version

| Attribute | Description |
| :-------- | :---------- |
| pgexporter_version | |

### pgexporter_postgresql_active

The state of PostgreSQL

| Attribute | Description |
| :-------- | :---------- |
| server | |

### pgexporter_postgresql_version

The PostgreSQL version

| Attribute | Description |
| :-------- | :---------- |
| server | |
| version | |
| minor_version | |
</details>

Are we looking for something like [this](https://stagb.in/c11yc4b)? (you may click on the M button on top right corner to view it in preview mode) @jesperpedersen

jesperpedersen:
@bassamadnan Yes, something along those lines.

The key is to tie it in with an external file that could have additional information about the metric in question

bassamadnan:
oh you mean another file altogether apart form the .md and .html generated having information about the metrics? @jesperpedersen 

Like this:

<details>
<summary>Additional info</summary>

# Prometheus Metrics Additional Information

## pgexporter_state

**Additional Information:** This metric indicates the overall state of the pgexporter system. A value of 1 means it is running normally.

**Usage Example:**
```
alert: pgexporter_state == 0
```

## pgexporter_postgresql_version

**Additional Information:** Shows the version of PostgreSQL that is being monitored. This includes attributes for the major version and minor version.

**Usage Example:**
```
dashboard: Track PostgreSQL version across your infrastructure
```
</details>

This could be refferenced somewhere in our md/html

jesperpedersen:
[...]

PR #217, opened 2025-04-30
[#213] Add bash script for documentation

bassamadnan:
Example Usage:

```bash
cd contrib/prometheus_scrape/
./prometheus.sh 5002 extra.info # port and path metric descriptions
```
Ensure +x permission for the script. Example output - [Link](https://stagb.in/fh9c0gi)

### DIFF
--- a/contrib/prometheus_scrape/prometheus.sh
+++ b/contrib/prometheus_scrape/prometheus.sh
@@ -1,0 +1,333 @@
+#!/bin/bash
+
+if [ $# -lt 2 ]; then
+    echo "Usage: $0 <port> <extra_info_file>"
+    exit 1
+fi
+
+PORT=$1
+EXTRA_INFO_FILE=$2
+TEMP_DIR="/tmp"
+METRICS_FILE="${TEMP_DIR}/prometheus.raw"
+EXTRA_INFO_PROCESSED="${TEMP_DIR}/extra.processed"
+MD_OUTPUT="prometheus.md"
+HTML_OUTPUT="prometheus.html"
+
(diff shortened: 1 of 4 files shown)


contrib/prometheus_scrape/prometheus.sh
```diff
@@ -0,0 +1,335 @@
...
+        db_value=""
+        if [[ "$line" =~ database=\"([^\"]+)\" ]]; then
+            db_value="${BASH_REMATCH[1]}"
+            if [[ "$db_value" == "postgres" ]]; then
+                line_priority=3
+            elif [[ "$db_value" == "test_db" ]]; then
+                line_priority=2
+            else
+                line_priority=1
+            fi
```
bassamadnan, line 293: output with "`postgres`" is selected by default, otherwise search for `test_db`  or just pick the first value after that.

jesperpedersen: What is `test_db` ? Remove that

bassamadnan: i use it for pgbench and was crosschecking the values. Will remove it

jesperpedersen:
@bassamadnan Please, add a README.md in the directory that explains the tool, and the format of `extra.info`.

I can see that you have a description of each of the metrics - hopefully that is optional since we can get the default description from the `#HELP` line

jesperpedersen:
@bassamadnan You need to add the files in the pgexporter.spec file

bassamadnan:
@jesperpedersen thanks, did not realize that. PTAL

Issue #224, opened 2025-05-25
Change path lengths


bassamadnan:
Several places in the codebase currently have `MISC_LENGTH` for files, ideally they should be `MAX_PATH`, we need to change this consistently (initialization, headers etc.) wherever files variables are concerned.

bassamadnan:
@jesperpedersen i'd like to work on this

jesperpedersen:
@bassamadnan Go for it !

PR #225, opened 2025-05-26
Use MAX_PATH for files

bassamadnan:
Addressing #224  I changed `pidfile` to MAX_PATH since that was done in pgagroal and pgmoneta as well, rest I think are understandable. @jesperpedersen  PTAL


jesperpedersen:
Merged.

PR #227, opened 2025-05-26
Use MAX_PATH for files

bassamadnan:
Changes missed in #225 and uncrustify


jesperpedersen:
Merged.

Issue #229, opened 2025-05-28
Extension Version Field


bassamadnan:
We need to build on top of the existing YAML format and add fields for extension version to load queries. This also means we need to implement a semantic versioning logic, maybe in a new `semvar.h` or in utils

jesperpedersen:
@bassamadnan I don't think we can use the existing YAML for this - so think about a new format that are specific for extensions - delta format, or all-in one

jesperpedersen:
@bassamadnan I don't think we need `tag` and `collector` for the extension format. We should just need a `metric` as the name for each query. We can get the prefix from the `struct extension`

bassamadnan:
> @bassamadnan I don't think we need `tag` and `collector` for the extension format. We should just need a `metric` as the name for each query. We can get the prefix from the `struct extension`

Yes, that makes sense

bassamadnan:
@jesperpedersen After recent discussions lets go with this format?
```yaml
extension: pg_stat_statements
metrics:
 - metric: queries
   queries:
     - query: SELECT queryid, calls FROM pg_stat_statements;
       version: "1.7"
       columns: [...]
     - query: SELECT queryid, calls, total_exec_time FROM pg_stat_statements;
       version: "1.8"
       columns: [...]
       
 - metric: database_stats
   queries:
     - query: SELECT dbname, calls FROM pg_stat_statements_db;
       version: "1.8"
       columns: [...]
```
Then in prometheus they would look like this 
```
pgexporter_pg_stat_statements_queries{...}
pgexporter_pg_stat_statements_database_stats{...}
```

PR #232, opened 2025-05-31
Add extension format

bassamadnan:
Added logic for loading extension config files, gave a barebones example. The way to load this is still the same (`-Y` flag). There is some duplicated code right now for AVL operations. PTAL @jesperpedersen @resyfer

### DIFF
--- a/extensions/CMakeLists.txt
+++ b/extensions/CMakeLists.txt
@@ -1,0 +1,39 @@
+#
+# Copyright (C) 2025 The pgexporter community
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list
+# of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this
+# list of conditions and the following disclaimer in the documentation and/or other
+# materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may
+# be used to endorse or promote products derived from this software without specific
+# prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+# all YAML files in current directory
+file(GLOB YAML_FILES "*.yaml" "*.yml")
+
+# Copy each YAML file to build directory
+foreach(YAML_FILE ${YAML_FILES})
+ get_filename_component(FILENAME ${YAML_FILE} NAME)
+ configure_file(${YAML_FILE} ${CMAKE_CURRENT_BINARY_DIR}/${FILENAME} COPYONLY)
+endforeach()
+
+# Install YAML files to system location
+install(FILES ${YAML_FILES} DESTINATION ${CMAKE_INSTALL_DATADIR}/pgexporter/extensions)
--- a/extensions/pg_stat_statements.yaml
+++ b/extensions/pg_stat_statements.yaml
@@ -1,0 +1,41 @@
+#
+# Copyright (C) 2025 The pgexporter community
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list
+# of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this
+# list of conditions and the following disclaimer in the documentation and/or other
+# materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may
+# be used to endorse or promote products derived from this software without specific
+# prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+# THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+# OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+extension: pg_stat_statements
+metrics:
+  - metric: queries
+    queries:
+      - query: SELECT queryid, calls, total_exec_time FROM pg_stat_statements;
+        version: "1.8"
+        columns:
+          - name: queryid
+            type: label
+          - name: calls
+            type: counter
+          - name: total_exec_time
+            type: counter
--- a/pgexporter.spec
+++ b/pgexporter.spec
@@ -36,6 +36,7 @@ cmake -DCMAKE_BUILD_TYPE=Release ..
 %{__mkdir} -p %{buildroot}%{_mandir}/man1
 %{__mkdir} -p %{buildroot}%{_mandir}/man5
 %{__mkdir} -p %{buildroot}%{_sysconfdir}/pgexporter
+%{__mkdir} -p %{buildroot}%{_datadir}/%{name}/extensions
 
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/LICENSE %{buildroot}%{_docdir}/%{name}/LICENSE
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/CODE_OF_CONDUCT.md %{buildroot}%{_docdir}/%{name}/CODE_OF_CONDUCT.md
@@ -49,7 +50,6 @@ cmake -DCMAKE_BUILD_TYPE=Release ..
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/doc/etc/pgexporter.service %{buildroot}%{_docdir}/%{name}/etc/pgexporter.service
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/shell_comp/pgexporter_comp.bash %{buildroot}%{_docdir}/%{name}/shell_comp/pgexporter_comp.bash
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/shell_comp/pgexporter_comp.zsh %{buildroot}%{_docdir}/%{name}/shell_comp/pgexporter_comp.zsh
-%{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/yaml/postgresql-12.yaml %{buildroot}%{_docdir}/%{name}/yaml/postgresql-12.yaml
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/yaml/postgresql-13.yaml %{buildroot}%{_docdir}/%{name}/yaml/postgresql-13.yaml
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/yaml/postgresql-14.yaml %{buildroot}%{_docdir}/%{name}/yaml/postgresql-14.yaml
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/yaml/postgresql-15.yaml %{buildroot}%{_docdir}/%{name}/yaml/postgresql-15.yaml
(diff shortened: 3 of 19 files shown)


src/libpgexporter/yaml_configuration.c
```diff
@@ -1195,11 +1252,171 @@ semantics_yaml(struct prometheus* prometheus, int prometheus_idx, yaml_config_t*
...
+
+         prom->pg_root = pgexporter_insert_node_avl(prom->pg_root, &new_query);
+      }
+   }
+
+   return 0;
+}
+
+static struct extension_metrics*
+find_or_create_extension(struct configuration* config, char* extension_name)
```
resyfer, line 1266: How about `search_or_add_extension`?

create made me initially think about CREATE EXTENSION, so I would say people _could_ get confused.

resyfer (changes requested):
Other than the requested changes, all LGTM.

src/include/utils.h
```diff
@@ -185,6 +185,27 @@ pgexporter_extract_message_from_data(char type, void* data, size_t data_size, st
...
 pgexporter_has_message(char type, void* data, size_t data_size);
 
+/**
+ * Parse a semantic version string (e.g., "1.8.2" or "2.1") into version struct
+ * @param version_str The version string to parse
+ * @param version Output version struct
+ * @return 0 on success, 1 on error
+ */
+int
+pgexporter_parse_extension_version(char* version_str, struct version* version);
```
jesperpedersen, line 195: They need to go to `extension.h`

jesperpedersen (changes requested):

jesperpedersen:
@bassamadnan Lets start with these... and you can't use the `-Y` parameter. The files has to be loaded from `share/pgexporter/` and in some way during a developer build. So, no CLI changes

jesperpedersen:
We can consider an `-E` override of a list of extensions... but that is a forum discussion

jesperpedersen:
This feature will require refactoring and we need a basic framework that can load `pg_stat_statements.yaml` automatically and enable it

bassamadnan:
@jesperpedersen Hi, the YAML flag is temporary to trigger the YAML APIs. We didnt have a discussion on it so I didnt add a new flag (how do we handle extensioms being enabled disabled etc since thats an entirely different thread.) Would you like to see a flag added with this patch?

jesperpedersen:
Yes, enable/disable and `-E` is for later. But, the `pgexporter/extensions` directory and its install is. So, the first merge will be quite large.

The extension files doesn't belong inside `contrib/` for sure

jesperpedersen:
@bassamadnan Just think about that f.ex. `query_alts.h` is the "base", `pg_query_alts.h` is specific to general PostgreSQL queries, and `ext_query_alts.h` is specific to extensions. The two latter uses the first as a base

src/libpgexporter/prometheus.c
```diff
@@ -1212,7 +1215,7 @@ extension_list_information(SSL* client_ssl, int client_fd)
          for (int i = 0; i < config->servers[server].number_of_extensions; i++)
          {
             safe_key1 = safe_prometheus_key(config->servers[server].extensions[i].name);
-            safe_key2 = safe_prometheus_key(config->servers[server].extensions[i].installed_version);
+            safe_key2 = safe_prometheus_key(config->servers[server].extensions[i].name);
```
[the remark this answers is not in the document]
bassamadnan: this was a mistake, I wanted to temporarily pass a `char*` argument inside this, installed version was changed to a struct of integers so I passed in something which is `char*`. Realised it that I had made a PR without changing this while I was about to sleep. My bad.

src/libpgexporter/yaml_configuration.c
bassamadnan: yup , makes sense. Will make this change

bassamadnan:
Yup, im working on it. @jesperpedersen  just to be sure, after these changes we will have a `XYZ/share/pgexporter/extensions` (containing YAML files for extensions) and existing `XYZ/share/doc/pgexporter/` ?

jesperpedersen:
@bassamadnan Yes, you can do an `share/pgexporter/extensions/` directory - try and build a RPM from the .spec file on how it works

bassamadnan:
> @bassamadnan Yes, you can do an `share/pgexporter/extensions/` directory - try and build a RPM from the .spec file on how it works

Sure, first time I'll be doing that -sorry if I make wrong commits, but will try my best

jesperpedersen:
@bassamadnan https://github.com/pgexporter/pgexporter/blob/main/doc/RPM.md should work

bassamadnan:
@jesperpedersen in the spec we try to install [this](https://github.com/pgexporter/pgexporter/blob/e8c7d12a8c8317de2565171239d9c0e6187f5ddd/pgexporter.spec#L52) yaml file, but it does not exist in the repositry (we start from -13.yaml), this is causing a failure in the build of an RPM package, im trying this through docker.

jesperpedersen:
@bassamadnan Yes, you can just remove that one

bassamadnan:
@jesperpedersen how will we get this extensions directory? just by reading the files in directory "../extensions/"? This will not be specified in cofig right? and since we are including it in our build it will work regardless

jesperpedersen:
@bassamadnan No, it isn't in the configuration.

Try to start with
```
char buf[MAX_PATH];
char* bin_path = realpath(argv[0], buf);
```

in `main.c`, and then check if it is a developer build or an installation. Then stick it into the main configuration

bassamadnan:
@jesperpedersen okay i was able to setup, but theres a slight caveat, this is how the code looks
```c
 *bin_path = realpath(argv0, NULL);
    if (*bin_path == NULL)
    {
        /* Can't resolve argv0, assume standard installation */
        snprintf(config->extensions_path, MAX_PATH, "/usr/local/share/pgexporter/extensions");
    }
    else if (strstr(*bin_path, "/build/src/") != NULL) { /* remove the executable name based on last slash and append ../extensions, this is for build/dev mode" */)
    else { /* undetermined path, return error */}
```
Is this how we are expecting it?, when i try it in install mode and run
```bash
pgexporter <other args>
```
the argv[0] resolves to "pgexporter" and relpath returns `NULL`

jesperpedersen:
@bassamadnan if it is `NULL` then `errno` should give you an idea ... but yes, along these lines

bassamadnan:
yes i confirmed, it returns error code 2 , in other words the `ENOENT` code. I think this will always be the case right? (if run from install, then realpath will always resolve to `NULL`) So we should do this instead:
1. Check if bin_path is not null AND if dev build (`/build/src`)  and manipulate to extension directory
2. Otherwise if bin path is null, assume the standard directory and check if we have read access to it
3. if we dont have read access to it (using `access()`) report error

jesperpedersen:
@bassamadnan Something like that - remember to clear `errno`

resyfer:
@bassamadnan please rebase, the main's updated since

bassamadnan:
> @bassamadnan please rebase, the main's updated since

yes done, weirdly it was an empty conflict? didn't think it would cause issues. PS - I typically check the comparision logic by adding these lines:
```c
printf("Comparing: %d.%d.%d to %d.%d.%d\n", temp->ext_version.major, temp->ext_version.minor, temp->ext_version.patch, ext_version->major, ext_version->minor, ext_version->patch);
```
inside the `while` loop of `pgexporter_get_extension_query_alt`
update 2: you dont need to specify `-Y` flag now, we compare names of the extension with the file name.

pgexporter.spec
```diff
@@ -49,7 +50,7 @@ cmake -DCMAKE_BUILD_TYPE=Release ..
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/doc/etc/pgexporter.service %{buildroot}%{_docdir}/%{name}/etc/pgexporter.service
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/shell_comp/pgexporter_comp.bash %{buildroot}%{_docdir}/%{name}/shell_comp/pgexporter_comp.bash
 %{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/shell_comp/pgexporter_comp.zsh %{buildroot}%{_docdir}/%{name}/shell_comp/pgexporter_comp.zsh
-%{__install} -m 644 %{_builddir}/%{name}-%{version}/contrib/yaml/postgresql-12.yaml %{buildroot}%{_docdir}/%{name}/yaml/postgresql-12.yaml
+%{__install} -m 644 %{_builddir}/%{name}-%{version}/build/extensions/pg_stat_statements.yaml %{buildroot}/usr/local/share/%{name}/extensions/
```
resyfer: Any specific reason to use `/usr/local/share/%{name}/...` instead of the usual `%{buildroot}%{_docdir}/%{name}`?

`%{_docdir}` itself is `/usr/share/doc`?

If this is a mistake, please try to use [RPM macros](https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/)...you can just take them from the neighbors of this line (and in other occurrences).

bassamadnan: We want to build it into share/pgexporter/extensiond and not inside doc. I think theres a difference in how RPM packaged it and wanted to ensure this directory exists regardless. I am not very happy with this myself and would like some suggestions

bassamadnan: This was being done since we hardcoded this path in the setup in main

jesperpedersen: @bassamadnan That is why you have to figure out if you are in `/usr/bin` or `/usr/local/bin` -- you can use `%{_datadir}` as the root for the .spec file

jesperpedersen: https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/

resyfer: What about this path:

```
%{buildroot}%{_docdir}/%{name}/yaml/extensions
```

?

If it's hard coded in `extension.c`, we can easily change it, so that's not really a problem I think, unless there is some other issues with doc?

bassamadnan: Okay thanks i'll take a look

bassamadnan: > What about this path:
> 
> ```
> %{buildroot}%{_docdir}/%{name}/yaml/extensions
> ```
> 
> ?
> 
> If it's hard coded in `extension.c`, we can easily change it, so that's not really a problem I think, unless there is some other issues?

Yes, i'll give it another proper look, the problem is some systems install it at other places, the one currently hardcoded does work for debian based systems. I'll add another check to confirm this as Jesper mentioned

jesperpedersen: it needs to go into `share/pgexporter/extensions` such that it follows the standard RPM way -- spend some time on it such you get it right

extensions/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# all YAML files in current directory
```
jesperpedersen, line 29: License header

extensions/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+file(GLOB YAML_FILES "*.yaml" "*.yml")
+
+# Copy each YAML file to build directory
+foreach(YAML_FILE ${YAML_FILES})
+    get_filename_component(FILENAME ${YAML_FILE} NAME)
```
jesperpedersen: Indentation

extensions/pg_stat_statements.yaml
```diff
@@ -0,0 +1,20 @@
+extension: pg_stat_statements
```
jesperpedersen, line 29: License header

extensions/pg_stat_statements.yaml
```diff
@@ -0,0 +1,20 @@
...
+metrics:
+  - metric: queries
+    queries:
+      - query: SELECT queryid, calls FROM pg_stat_statements;
+        version: "1.7"
```
jesperpedersen: We need to support 1.8 to 1.12

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <pgexporter.h>
+#include <extension.h>
+#include <ext_query_alts.h>
+#include <shmem.h>
+
+// Get height of extension AVL Tree Node
+static int ext_height(struct ext_query_alts* A);
```
jesperpedersen: No capitals as variable names

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+// Right rotate extension node
+static struct ext_query_alts* ext_node_right_rotate(struct ext_query_alts* root);
+
+// Left rotate extension node
+static struct ext_query_alts* ext_node_left_rotate(struct ext_query_alts* root);
+
+void
+pgexporter_copy_extension_query_alts(struct ext_query_alts** dst, struct ext_query_alts* src)
+{
+   if (!src)
```
jesperpedersen, line 52: Assign a local variable first, and it would be src, dst -- output is last

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+   {
+      return;
+   }
+
+   void* new_query_alt = NULL;
```
jesperpedersen: Variable def in the beginning of the function

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+   (*dst)->ext_version = src->ext_version;
+   (*dst)->node.is_histogram = src->node.is_histogram;
+   (*dst)->node.n_columns = src->node.n_columns;
+
+   memcpy((*dst)->node.query, src->node.query, MAX_QUERY_LENGTH);
+   memcpy((*dst)->node.columns, src->node.columns, MAX_NUMBER_OF_COLUMNS * sizeof(struct column));
+
+   pgexporter_copy_extension_query_alts(&(*dst)->left, src->left);
+   pgexporter_copy_extension_query_alts(&(*dst)->right, src->right);
+}
```
jesperpedersen, line 72: Assign output here

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+
+static int
+ext_height(struct ext_query_alts* A)
```
jesperpedersen: Same

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+{
+   return A ? A->height : 0;
+}
+
+static int
+ext_get_node_balance(struct ext_query_alts* A)
```
jesperpedersen: Create meaningful variable names

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+
+static struct ext_query_alts*
+ext_node_left_rotate(struct ext_query_alts* root)
+{
+   if (!root || !root->right)
+   {
+      return root;
+   }
+
+   struct ext_query_alts* A = root;
```
jesperpedersen: Move

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+
+struct ext_query_alts*
+pgexporter_insert_extension_node_avl(struct ext_query_alts* root, struct ext_query_alts** new_node)
+{
+   if (!root)
+   {
+      return (*new_node);
+   }
+
+   int cmp = pgexporter_compare_extension_versions(&root->ext_version, &(*new_node)->ext_version);
```
jesperpedersen: Variable defs first

src/libpgexporter/ext_query_alts.c
```diff
@@ -0,0 +1,238 @@
...
+struct ext_query_alts*
+pgexporter_get_extension_query_alt(struct ext_query_alts* root, struct version* ext_version)
+{
+   struct ext_query_alts* temp = root;
+   struct ext_query_alts* last = NULL;
+
+   // Traversing the AVL tree to find highest compatible version
+   while (temp)
+   {
+      int cmp = pgexporter_compare_extension_versions(&temp->ext_version, ext_version);
```
jesperpedersen, line 194: This is ok

src/libpgexporter/extension.c
```diff
@@ -0,0 +1,364 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+int
+pgexporter_setup_extensions_path(struct configuration* config, const char* argv0, char** bin_path)
+{
+   errno = 0;
+
+   *bin_path = realpath(argv0, NULL);
```
jesperpedersen: Local variable first, then assign

jesperpedersen:
@bassamadnan Lets start with these first

extensions/pg_stat_statements.yaml
bassamadnan: Hi @jesperpedersen, we will be supporting them yes. But i'll make a seperate PR with exhaustive queries for pg stat. The idea with this file is to show that the current AVL setup works and this extensiom file is parsed and loaded sucessfuly.

extensions/CMakeLists.txt
bassamadnan: done, thanks! May I please know why we are including license only for this CMakeLists?

extensions/pg_stat_statements.yaml
jesperpedersen: Well, just remove the 1.7 query then - PostgreSQL 13 has 1.8

extensions/CMakeLists.txt
jesperpedersen: All files should have a license header

extensions/pg_stat_statements.yaml
bassamadnan: Okay, thanks

jesperpedersen:
Merged - although I'm not 100% on this - we will find the proper split later I think.

bassamadnan:
@jesperpedersen Thanks, were you thinking of something else or should I keep something else in mind? (For the record, this addresses #229).

jesperpedersen:
@bassamadnan We will see about the requirements once we get more extensions supported - for now, just keep working in this direction

PR #236, opened 2025-06-05
Remove extra prefix

bassamadnan:
Fixes extension queries metric by removing the extra `pgexporter_` in its prefix.


jesperpedersen:
Merged.

Issue #237, opened 2025-06-06
Memory leak in pgexporter-admin


resyfer:
**Describe the bug**

Title

**To Reproduce**

Already one user added, then tried to add another. This user had a password with length 1 character, and it asked me for password and verify twice.

**Version**

0.7.0

**PostgreSQL**

17

**Debug logs**

```
pgexporter-admin -f pgexporter_users.conf user add
User name: resyfer
Password :
Verify   :
Password :
Verify   :
Header:
  ClientVersion: 0.7.0
  Command: 12
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20250606162943
Outcome:
  Status: true
  Time: 00:00:05
Response:
  Users:
    - pgexporter
    - resyfer

=================================================================
==1729==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 6 byte(s) in 1 object(s) allocated from:
    #0 0x7f650cae6f0b in malloc (/lib64/libasan.so.8+0xe6f0b) (BuildId: 7f1aa7e2e600e8c9d54ce6e3d36f3d31bfe7949a)
    #1 0x7f650bebe02f in pgexporter_get_password /home/resyfer/pgexporter/src/libpgexporter/utils.c:799
    #2 0x562e0f80a49b in add_user /home/resyfer/pgexporter/src/admin.c:664
    #3 0x562e0f808de2 in main /home/resyfer/pgexporter/src/admin.c:281
    #4 0x7f650b4115f4 in __libc_start_call_main (/lib64/libc.so.6+0x35f4) (BuildId: 2b3c02fe7e4d3811767175b6f323692a10a4e116)
    #5 0x7f650b4116a7 in __libc_start_main@@GLIBC_2.34 (/lib64/libc.so.6+0x36a7) (BuildId: 2b3c02fe7e4d3811767175b6f323692a10a4e116)
    #6 0x562e0f807884 in _start (/usr/local/bin/pgexporter-admin+0x884) (BuildId: 0b41b1e42d7f423b75e61e267fdd1e1407aaa9a5)

SUMMARY: AddressSanitizer: 6 byte(s) leaked in 1 allocation(s).
```

jesperpedersen:
Luca did some changes in https://github.com/agroal/pgagroal/commit/af9feff09f5955765b5b5e0e44ac3311f7b66a57 too

Userfrom1995:
I believe this commit fixes the issue:  
[...]

PR #238, opened 2025-06-06
[#237] Fix Memory leak in pgexporter-admin

Userfrom1995:


jesperpedersen:
@Userfrom1995 You can port Luca's changes as well - https://github.com/agroal/pgagroal/commit/af9feff09f5955765b5b5e0e44ac3311f7b66a57

src/admin.c
```diff
@@ -676,6 +676,7 @@ add_user(char* users_path, char* username, char* password, bool generate_pwd, in
    {
       if ((unsigned char)(*(password + i)) & 0x80)
       {
+         warnx("Illegal character(s) in password");
```
jesperpedersen, line 679: Don't we need the free/NULL here too ?

src/admin.c
```diff
@@ -911,6 +915,7 @@ update_user(char* users_path, char* username, char* password, bool generate_pwd,
          {
             if ((unsigned char)(*(password + i)) & 0x80)
             {
+               warnx("Illegal character(s) in password");
```
jesperpedersen, line 922: Same ?

src/admin.c
Userfrom1995: Yes, it's needed. I just tested it.

src/admin.c
Userfrom1995: Yes.

jesperpedersen:
Merged.

jesperpedersen:
@Userfrom1995 Please, make sure that things are unified acroos the projects

Issue #241, opened 2025-06-13
Fix memory leaks for HTTP / Bridge / extension usage


jesperpedersen:
There are memory on both successful and error paths

bassamadnan:
Can we also rename this to include bridge? I'll open a PR with this issue number

jesperpedersen:
@bassamadnan Of course

bassamadnan:
> [@bassamadnan](https://github.com/bassamadnan) Of course

Thank you

bassamadnan:
@jesperpedersen For some reason (only on GitHub Actions) we encounter this issue on one of the test cases:

<details>
<summary>AddressSanitizer Error Log</summary>

```
==4256==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x52900000e200 at pc 0x0000002bd909 bp 0x7ffffffbc110 sp 0x7ffffffbb8c8
READ of size 4 at 0x52900000e200 thread T0
    #0 0x2bd908 in __asan_memcpy /usr/src/contrib/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:63:3
    #1 0x800b888a3 in pgexporter_read_int32 /home/runner/work/pgexporter/pgexporter/src/libpgexporter/utils.c:351:4
    #2 0x800b8af98 in pgexporter_has_message /home/runner/work/pgexporter/pgexporter/src/libpgexporter/utils.c:243:20
    #3 0x800b6c75f in query_execute /home/runner/work/pgexporter/pgexporter/src/libpgexporter/queries.c:648:17
    #4 0x800b6e94a in pgexporter_query_settings /home/runner/work/pgexporter/pgexporter/src/libpgexporter/queries.c:419:11
    #5 0x800b46309 in settings_information /home/runner/work/pgexporter/pgexporter/src/libpgexporter/prometheus.c:1470:16
    #6 0x800b40992 in metrics_page /home/runner/work/pgexporter/pgexporter/src/libpgexporter/prometheus.c:689:10
    #7 0x800b3e9b8 in pgexporter_prometheus /home/runner/work/pgexporter/pgexporter/src/libpgexporter/prometheus.c:242:7
    #8 0x304e94 in accept_metrics_cb /home/runner/work/pgexporter/pgexporter/src/main.c:1538:7
    #9 0x8003549df in ev_invoke_pending (/usr/local/lib/libev.so.4+0x79df)
[...]

PR #246, opened 2025-06-15
Use malloc in http_get

bassamadnan:
Revamped the `http_get` to match exactly the one we use in `pgmoneta`. This was missing some validation checks earlier since we were receiving corrupted data from the bridge endpoint, now that the bridge issue has been fixed this free's the allocated memory on all paths. 

Apart from that also added a debug statement incase we try to read an integer out of bounds. Without this check, `curl` fails on FreeBSD debug mode on CI.
Part of the ongoing fixes in #241

### DIFF
--- a/src/libpgexporter/http.c
+++ b/src/libpgexporter/http.c
@@ -168,24 +168,23 @@ error:
 int
 pgexporter_http_get(struct http* http, char* hostname, char* path)
 {
-   struct message msg_request;
-   struct message* msg_response = NULL;
+   struct message* msg_request = NULL;
    int error = 0;
    int status;
    char* request = NULL;
    char* full_request = NULL;
    char* response = NULL;
    char* user_agent = NULL;
-   char* endpoint = (path != NULL) ? path : "/metrics";
-
-   memset(&msg_request, 0, sizeof(struct message));
 
    pgexporter_log_trace("Starting pgexporter_http_get");
-   if (http_build_header(PGEXPORTER_HTTP_GET, endpoint, &request))
+   if (path == NULL || strlen(path) == 0)
    {
-      pgexporter_log_error("Failed to build HTTP header");
+      pgexporter_log_error("GET: Path can not be NULL");
       goto error;
    }
+   request = pgexporter_append(request, "GET ");
+   request = pgexporter_append(request, path);
+   request = pgexporter_append(request, " HTTP/1.1\r\n");
 
    pgexporter_http_add_header(http, "Host", hostname);
    user_agent = pgexporter_append(user_agent, "pgexporter/");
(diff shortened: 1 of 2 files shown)


src/libpgexporter/http.c
```diff
@@ -168,24 +168,20 @@ http_extract_headers_body(char* response, struct http* http)
...
    int error = 0;
    int status;
    char* request = NULL;
    char* full_request = NULL;
    char* response = NULL;
    char* user_agent = NULL;
-   char* endpoint = (path != NULL) ? path : "/metrics";
-
-   memset(&msg_request, 0, sizeof(struct message));
+   char* endpoint = path ? path : "/get";
```
jesperpedersen: I don't think we want this - path is the path; there is no implicit

jesperpedersen (changes requested):

src/libpgexporter/http.c
bassamadnan: oh right, i forgot we already had this discussion [here](https://github.com/pgexporter/pgexporter/pull/222#discussion_r2106192986) I'll revert

bassamadnan: actually i'll just add an error so that its always specified i think, we already specify it everywhere anyway.

src/libpgexporter/http.c
```diff
@@ -168,24 +168,23 @@ http_extract_headers_body(char* response, struct http* http)
...
    char* full_request = NULL;
    char* response = NULL;
    char* user_agent = NULL;
-   char* endpoint = (path != NULL) ? path : "/metrics";
-
-   memset(&msg_request, 0, sizeof(struct message));
 
    pgexporter_log_trace("Starting pgexporter_http_get");
-   if (http_build_header(PGEXPORTER_HTTP_GET, endpoint, &request))
+   if (path == NULL)
```
jesperpedersen: Nor empty

src/libpgexporter/http.c
```diff
@@ -168,24 +168,23 @@ http_extract_headers_body(char* response, struct http* http)
...
    {
-      pgexporter_log_error("Failed to build HTTP header");
+      pgexporter_log_error("Path can not be NULL");
```
jesperpedersen: Prefix with GET:

src/libpgexporter/http.c
```diff
@@ -219,31 +226,12 @@ pgexporter_http_get(struct http* http, char* hostname, char* path)
...
-         goto error;
-      }
-   }
+   status = pgexporter_http_read(http->ssl, http->socket, &response);
 
-   if (msg_response != NULL && msg_response->length > 0 && msg_response->data != NULL)
+   if (response == NULL)
    {
-      response = pgexporter_append(response, (char*)msg_response->data);
+      pgexporter_log_error("No response data collected");
```
jesperpedersen: Prefix with GET:

jesperpedersen (changes requested):

jesperpedersen:
Merged.

jesperpedersen:
@bassamadnan Please, port

bassamadnan:
> @bassamadnan Please, port

Yes will do.

Issue #235, opened 2025-06-05
Create a basic test suite


jesperpedersen:
Create a basic test suite that verifies that certain Prometheus metrics are in the output

jesperpedersen:
@bassamadnan Please, take the this one

jesperpedersen:
@bassamadnan Port from pgmoneta, and use the HTTP API to verify that the core metrics exists

jesperpedersen:
@ashu3103 Please, help @bassamadnan if he needs it

bassamadnan:
@jesperpedersen Hapilly will. But just to be sure, we can only test the HTTP API via a bridge, by curling the bridge endpoint I mean. And for now should I add test cases on Release mode only? Because on DEBUG we will detect memory leaks, although i can just ignore this and verify that HTTP API atleast works.

jesperpedersen:
@bassamadnan Thanks, we can fail the build if there are sanizer failures. For now, follow pgmoneta in this, and coordinate with @ashu3103

bassamadnan:
@ashu3103 @jesperpedersen ok so the steps will be
1. Add the test suite from pgmoneta, initialize cluster with pgmoneta settings
2. This means creating a `pg_monitor` user and creating a `_.conf` with bridge enabled for `pgexporter`, much of this can be taken from `pgmoneta`
3. TC # 1 -> Curl through pgexporter directly, bare bones setup test
4. TC # 2 -> Curl through the bridge, this checks the HTTP APIs
5.  Clean all of this, again taken from `pgmoneta`

@ashu3103  - I have added you as a collaborator feel free to step in wherever you wish to, regardless i'll reach out to you to understand any flows from pgmoneta.

jesperpedersen:
@bassamadnan The idea is to use `http.h` instead of curl such we can find requirements

bassamadnan:
Oh right, i forgot we could do that from the testsuite library we use. @jesperpedersen 

Will do that. Will make a call if we have to pattern match or do something else to confirm the metrics from the response.

jesperpedersen:
@bassamadnan I think we can iterate through each line in the body and do `pgexporter_starts_with()` and `pgexporter_ends_with()` for now. Stick that in a function, and then we can test against core and the extensions installed

bassamadnan:
[...]

PR #247, opened 2025-06-15
Add test suite

bassamadnan:
Addressing #235, Heres what is being done in the testsuite script (`.sh`):

1. setup pgexporter with configuration required and appropriate roles
2. `curl` and log the response to `:5002/metrics` and `:5003/metrics` (bridge)

Then in the test suite we have this:
1. test cases for status, ping to check if successful startup
2. check connection to postgres database, get the version query and validate it against the one installed in CI, check if `extension_path` is detected (not null/empty)
3. use `http_get` for requesting metrics on `5002/metrics` and `5003/metrics` and also validating for expected response by matching first few lines. Also detects if extension `pg_stat_statements` has been detected (it is installed in the script to the postgres database) before shutting down

We also save the logs as an artifact to check for issues incase something fails

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
         build_type: [Debug, Release]
       fail-fast: false
 
-
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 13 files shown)


bassamadnan:
Without the last few PRs - something or the other would fail, this can be confirmed by reverting the code and running these tests. Alot of bugs had been unearthed while developing these

bassamadnan:
@jesperpedersen @resyfer PTAL whenever you get some time, thanks!

jesperpedersen:
@bassamadnan testsuite.sh need a 0755 execution mask.

When I did
```
. ./testsuite.sh
```

the terminal disappeared

bassamadnan:
> @bassamadnan testsuite.sh need a 0755 execution mask.
> 
> When I did
> 
> ```
> . ./testsuite.sh
> ```
> 
> the terminal disappeared

I have given it execution permision, could you try now please

jesperpedersen:
```
=== DEBUG: Creating master key ===
pgexporter-admin: The file ~/.pgexporter/master.key already exists
Header:
  ClientVersion: 0.7.0
  Command: 11
  Compression: 0
  Encryption: 0
  Output: 0
  Timestamp: 20250617072338
Outcome:
  Error: 1
  Status: false
pgexporter-admin: Cannot generate master key
```

This should not be a failure

bassamadnan:
> ```
> === DEBUG: Creating master key ===
> pgexporter-admin: The file ~/.pgexporter/master.key already exists
> Header:
>   ClientVersion: 0.7.0
>   Command: 11
>   Compression: 0
>   Encryption: 0
>   Output: 0
>   Timestamp: 20250617072338
> Outcome:
>   Error: 1
>   Status: false
> pgexporter-admin: Cannot generate master key
> ```
> 
> This should not be a failure

Okay i have added a `|| true` , I always run it on actions so never ran into these issues. Tested it locally now

jesperpedersen:
Merged.

Issue #248, opened 2025-06-18
Add support for pg_stat_statements


bassamadnan:
**Is your feature request related to a problem? Please describe.**

Yes, we need to support multiple versions of [pg_stat_statements](https://www.postgresql.org/docs/current/pgstatstatements.html)

**Describe the solution you'd like**
Queries in the `extensions/` with support for new features added across vesions

**What is the SQL statement you would like to support**
SQL queries must be fitting with utilities that `pg_stat_statements` provide and should be relevant to `pgexporter`

PR #249, opened 2025-06-22
Add queries for pg_stat_statements

bassamadnan:
Addressing #248 
Added comprehensive metrics coverage for pg_stat_statements extension across versions from 1.8 (Default version of the minimum postgres version supported by `pgexporter`):

Heres what majorly changed across them:

**Version 1.8**: Planning statistics, WAL metrics, execution timing separation  
**Version 1.9**: Query hierarchy (toplevel field), pg_stat_statements_info view  
**Version 1.10**: JIT compilation metrics, separate temp block I/O timing  
**Version 1.11**: Detailed I/O timing by storage type, JIT deform operations, statistics reset tracking

These are majorly what I found relevant for pgexporter, some help/advice from AI as well and comments on the core postgres pg_stat_statements for motivation of updates.  I have also documented the diffs accross the versions [here](https://github.com/bassamadnan/pg_ext_tracker/tree/main/pg_stat_statements/docker/pg_ext_diffs).

PTAL @jesperpedersen @resyfer

### DIFF
--- a/extensions/pg_stat_statements.yaml
+++ b/extensions/pg_stat_statements.yaml
@@ -28,14 +28,438 @@
 
 extension: pg_stat_statements
 metrics:
-  - metric: queries
+
+#
+# Extension Version 1.8
+# Added: Planning statistics, WAL metrics, execution timing separation
+#
+
+  # Most executed queries - monitoring query frequency
+  - metric: most_executed
     queries:
-      - query: SELECT queryid, calls, total_exec_time FROM pg_stat_statements;
+      - query: SELECT calls, query FROM pg_stat_statements ORDER BY calls DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: calls
+            type: gauge
+            description: Number of times the SQL query is executed
+          - name: query
+            type: label
+
+  # Most planned queries
+  - metric: most_planned
+    queries:
+      - query: SELECT plans, query FROM pg_stat_statements ORDER BY plans DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: plans
+            type: gauge
+            description: Number of times the sql query is planned
+          - name: query
+            type: label
+
+  # Most row-affecting queries - data volume impact
+  - metric: most_rows
+    queries:
+      - query: SELECT rows, query FROM pg_stat_statements ORDER BY rows DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: rows
+            type: gauge
+            description: Number of rows the SQL query affects
+          - name: query
+            type: label
+
+  # Slowest queries by execution time
+  - metric: slowest_execution
+    queries:
+      - query: SELECT total_exec_time, query FROM pg_stat_statements ORDER BY total_exec_time DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: total_exec_time
+            type: gauge
+            description: Milliseconds taken by the sql query to execute
+          - name: query
+            type: label
+
+  # Highest WAL generating queries
+  - metric: highest_wal
+    queries:
+      - query: SELECT wal_bytes, query FROM pg_stat_statements ORDER BY wal_bytes DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: wal_bytes
+            type: gauge
+            description: Bytes occupied in WAL
+          - name: query
+            type: label
+
+  # Planning statistics with detailed metrics - query optimization insights
+  - metric: planning_stats
+    queries:
+      - query: SELECT queryid, plans, total_plan_time, mean_plan_time, query FROM pg_stat_statements WHERE plans > 0 ORDER BY total_plan_time DESC LIMIT 10;
         version: "1.8"
         columns:
           - name: queryid
             type: label
-          - name: calls
-            type: counter
+          - name: plans
+            type: gauge
+            description: Number of times the query has been planned
+          - name: total_plan_time
+            type: gauge
+            description: Total planning time in milliseconds
+          - name: mean_plan_time
+            type: gauge
+            description: Mean planning time in milliseconds
+          - name: query
+            type: label
+
+  # Execution time with detailed metrics - identify performance bottlenecks
+  - metric: exec_time_detailed
+    queries:
+      - query: SELECT queryid, total_exec_time, mean_exec_time, query FROM pg_stat_statements ORDER BY total_exec_time DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: queryid
+            type: label
           - name: total_exec_time
-            type: counter
+            type: gauge
+            description: Total execution time in milliseconds
+          - name: mean_exec_time
+            type: gauge
+            description: Mean execution time in milliseconds
+          - name: query
+            type: label
+
+  # I/O intensive queries - disk access patterns
+  - metric: io_stats
+    queries:
+      - query: SELECT queryid, shared_blks_hit, shared_blks_read, shared_blks_written, local_blks_read, local_blks_written, temp_blks_read, temp_blks_written, query FROM pg_stat_statements WHERE shared_blks_read + local_blks_read + temp_blks_read > 0 ORDER BY (shared_blks_read + local_blks_read + temp_blks_read) DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: queryid
+            type: label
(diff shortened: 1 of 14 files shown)


bassamadnan:
I think the documentation is failing to build on FreeBSD ... ?

bassamadnan:
Okay added `dot -c` , copied from the recent changes in [pgagroal](https://github.com/agroal/pgagroal/commit/5909ce241838a6168abd180b93235cbec06cb840)

jesperpedersen:
```
pgexporter_pg_stat_statements_io_stats_local_blks_read{server="primary",queryid="-5746151688168999853",query="SELECT
  d_datname as \"Name\",
  pg_catalog_pg_get_userbyid(d_datdba) as \"Owner\",
  pg_catalog_pg_encoding_to_char(d_encoding) as \"Encoding\",
  CASE d_datlocprovider WHEN $1 THEN $2 WHEN $3 THEN $4 WHEN $5 THEN $6 END AS \"Locale Provider\",
  d_datcollate as \"Collate\",
  d_datctype as \"Ctype\",
  d_datlocale as \"Locale\",
  d_daticurules as \"ICU Rules\",
  CASE WHEN pg_catalog_array_length(d_datacl, $7) = $8 THEN $9 ELSE pg_catalog_array_to_string(d_datacl, $10) END AS \"Access privileges\"
FROM pg_catalog_pg_database d
ORDER BY 1"} 0
```

Did you test this in a Prometheus frontend, like Grafana ? I mean the "as" statements...

Remember that there is the `pg_stat_statements_reset()` so `gauge` is safer than `counter`

jesperpedersen:
@bassamadnan After `extension.c:354` we need an INFO after that says something like

```
INFO Extension my_extension not enabled for metrics
```

bassamadnan:
> 
> Did you test this in a Prometheus frontend, like Grafana ? I mean the "as" statements...
> 
> Remember that there is the `pg_stat_statements_reset()` so `gauge` is safer than `counter`

No, i did not test this, i assumed since we already have `pgexporter_pg_stat_statements_plans` in our core metrics (as a counter that too) this would be fine as well, but i'll check it out and update you on this.

1. On that note any pointers on when to keep a metric as `gauge`, `counter` etc.. I remember seeing a discussion around this between you and @resyfer but I cant find it :(
2. Since we are `pg_stat_statements` in the core as an extension, should we remove their queries from  `internal.h` (or maybe even `contrib/yaml` or `contrib/json`) and add keep them only here ?

jesperpedersen:
`gauge` can go up and down, `counter` can only go up.

Yes, `pg_stat_statements` should only be supported through the extension work, not pgexporter/core

bassamadnan:
> `gauge` can go up and down, `counter` can only go up.
> 
> Yes, `pg_stat_statements` should only be supported through the extension work, not pgexporter/core

Understood, then with this PR I will make those changes as well. And yes this looks fine in prometheus dashboard (nothing is breaking) Heres a quick look up [link](https://pasteboard.co/fwS9Eu3SBAA0.png) to the metric u mentioned in prometheus dashboard. Will change to gauge wherever appropriate, i'll need to look up what can the reset function reset

jesperpedersen:
@bassamadnan Just do it in this PR. Then I think `pg_stat_statements` is done

bassamadnan:
> @bassamadnan Just do it in this PR. Then I think `pg_stat_statements` is done

Yes yes, with this PR

update: [pg_stat_statements_reset](https://www.postgresql.org/docs/13/pgstatstatements.html#id-1.11.7.38.7.2.1.1.2) resets everything apparently, so its fair to change it all to `gauge`

jesperpedersen:
@bassamadnan You can look at contrib/grafana/postgresql_dashboard_*.json as part of this too... or as a separate PR.

We have to figure out how to bring extensions into the Grafana dashboard

bassamadnan:
> @bassamadnan You can look at contrib/grafana/postgresql_dashboard_*.json as part of this too... or as a separate PR.
> 
> We have to figure out how to bring extensions into the Grafana dashboard


Sure, but i think we should ponder on it abit later. Extensions are abit tricky to handle with these dashboards since we might have to create seperate files for extension versions otherwise we will see a blank visualization (same issue as postgres versions). On that note i'll try to include the core metrics in the extension yaml with the same name (so existing dashboards work just fine) as they have right now, fortunately the `pg_stat_statements` metrics start with the extension name, same as our implementation for extensions.

I dont mind creating seperate files in `contrib/grafana/extensions/pg_stat_statements<version>.json` though, but i'll see if theres a better way

jesperpedersen:
@bassamadnan You can just change the existing Grafana files to use the new names...

I don't think we need JSON support for extension work - if that doesn't work out then we should do a yaml2json in contrib/

bassamadnan:
> @bassamadnan You can just change the existing Grafana files to use the new names...

Right, thats simpler. My bad.

> I don't think we need JSON support for extension work - if that doesn't work out then we should do a yaml2json in contrib/

No i was talking about Grafana dashboards for the extension,they have a different format altogether (through JSON, which has to be imported in Grafana to view the dashboard). I was not talking about our extension support through `pgexporter` that will stay in YAML only.

jesperpedersen:
@bassamadnan Once we have #250 it will be easier to test that things are working

bassamadnan:
> @bassamadnan After `extension.c:354` we need an INFO after that says something like
> 
> ```
> INFO Extension my_extension not enabled for metrics
> ```

We already have a debug line in `:314` I think I'll change that to info, that checks if its enabled, so we never reach `:354`

bassamadnan:
> @bassamadnan Once we have #250 it will be easier to test that things are working

That is a very nice issue

bassamadnan:
I honestly dont know why/how changing any of these files would stop `pgexporter` from starting @jesperpedersen . In the CI, pgexporter fails to start for some reason. I dont know why

Start ups just fine on my freebsd vm

extensions/pg_stat_statements.yaml
```diff
@@ -28,14 +28,447 @@
...
+    queries:
+      - query: SELECT wal_bytes, query FROM pg_stat_statements ORDER BY wal_bytes DESC LIMIT 10;
+        version: "1.8"
+        columns:
+          - name: wal_bytes
+            type: gauge
+            description: Bytes occupied in WAL
+          - name: query
+            type: label
+
```
Userfrom1995, line 96: @bassamadnan try removing this duplicate query [line 96-104].

jesperpedersen: There should be an ERROR for duplicated entries

bassamadnan: > @bassamadnan try removing this duplicate query [line 96-104].

Okay that helped, that really was the issue. Thanks!

bassamadnan:
@jesperpedersen PTAL

src/libpgexporter/extension.c
```diff
@@ -311,8 +311,8 @@ pgexporter_load_extension_yamls(struct configuration* config)
          }
          else
          {
-            pgexporter_log_debug("Skipping disabled extension: %s",
-                                 config->servers[server].extensions[i].name);
+            pgexporter_log_info("Extension %s not enabled for metrics on: %s",
+                                config->servers[server].name, config->servers[server].extensions[i].name);
```
jesperpedersen: Wrong order

jesperpedersen:
@bassamadnan You are missing the check for duplicate entries

bassamadnan:
> @bassamadnan You are missing the check for duplicate entries

That would be a new issue no? I am thinking about keeping  a trie data structure which keeps track of metirc names across all the AVL query nodes which we have, that way we will determine if a metric already exists

src/libpgexporter/extension.c
bassamadnan: `main` -> `pgexporter_load_extension_yamls` -> for each server check if its enabled first before calling `pgexporter_load_single_extension_yaml` 

In multiple endpoints/servers we may have different extensions enabled/disabled, are you thinking of some other order?

jesperpedersen:
@bassamadnan You can create a separate issue as well

src/libpgexporter/extension.c
jesperpedersen: It prints the server name as the first string, and extension name as the second -- it should be the other way around

bassamadnan: oh my bad

jesperpedersen:
Merged.

PR #257, opened 2025-06-28
Detect duplicate metric names

bassamadnan:
Added logic for detecting duplicate metric names, this includes handling different queries based on the same version, validating before processing the file and maintaining a global (`configuration :: metric_name_trie`) list of metrics processed.

Also with this PR we can only read either the YAML/JSON files, or the internal metrics (not both together). Also changed `pg_db_conn` to `pg_db_conn_ssl` to resolve the name conflict.

### DIFF
--- a/contrib/json/postgresql-13.json
+++ b/contrib/json/postgresql-13.json
@@ -439,7 +439,7 @@
           ]
         }
       ],
-      "tag": "pg_db_conn",
+      "tag": "pg_db_conn_ssl",
       "sort": "data",
       "collector": "connections"
     },
--- a/contrib/json/postgresql-14.json
+++ b/contrib/json/postgresql-14.json
@@ -439,7 +439,7 @@
           ]
         }
       ],
-      "tag": "pg_db_conn",
+      "tag": "pg_db_conn_ssl",
       "sort": "data",
       "collector": "connections"
     },
--- a/contrib/json/postgresql-15.json
+++ b/contrib/json/postgresql-15.json
@@ -439,7 +439,7 @@
           ]
         }
       ],
-      "tag": "pg_db_conn",
+      "tag": "pg_db_conn_ssl",
       "sort": "data",
       "collector": "connections"
     },
--- a/contrib/json/postgresql-16.json
+++ b/contrib/json/postgresql-16.json
@@ -397,7 +397,7 @@
           ]
         }
       ],
-      "tag": "pg_db_conn",
+      "tag": "pg_db_conn_ssl",
       "sort": "data",
       "collector": "connections"
     },
--- a/contrib/json/postgresql-17.json
+++ b/contrib/json/postgresql-17.json
@@ -362,7 +362,7 @@
           ]
         }
       ],
-      "tag": "pg_db_conn",
+      "tag": "pg_db_conn_ssl",
       "sort": "data",
       "collector": "connections"
     },
--- a/contrib/yaml/postgresql-13.yaml
+++ b/contrib/yaml/postgresql-13.yaml
@@ -416,7 +416,7 @@ metrics:
           type: label
         - type: gauge
           description: Number of DB connections with SSL.
-    tag: pg_db_conn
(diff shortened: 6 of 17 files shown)


jesperpedersen:
@bassamadnan Can't we use ART instead of trie ?
@Jubilee101 Cc:

bassamadnan:
> @bassamadnan Can't we use ART instead of trie ? @Jubilee101 Cc:

I suppose we could, but i have never worked with ART before, and since this is just being done during startup i wanted to keep things simple (and easier to debug).

jesperpedersen:
@bassamadnan But, trie would be a core API - take a look at `prometheus_client.c` for an example

bassamadnan:
> @bassamadnan But, trie would be a core API - take a look at `prometheus_client.c` for an example

Okay, I can try using ART instead. The existing API usage in `prometheus_client` do help, `art.h` has a `MAX_PREFIX_LEN` set to 10 right now. I think that should change, also some logic to validate a metric with valid characters is also needed (like `char_to_index` in `trie.c` in this PR)

jesperpedersen:
https://db.in.tum.de/~leis/papers/ART.pdf

Jubilee101:
> I suppose we could, but i have never worked with ART before, and since this is just being done during startup i wanted to keep things simple (and easier to debug).

Think of it as an enhanced Trie, with much better space efficiency. And we already have APIs, you can use them directly.

> But, trie would be a core API - take a look at prometheus_client.c for an example

Not sure if we need it in core, ART can handle what trie does, they are basically the same idea.

> The existing API usage in prometheus_client do help, art.h has a MAX_PREFIX_LEN set to 10 right now. I think that should change

I don't understand the concern here, can you elaborate? `MAX_PREFIX_LEN` is the size of prefix "cache", we don't save the rest of the prefix if it's longer than that. Having the first 10 chars of a prefix serves as a filter for those whose prefixes diverge before the first 10 characters. If you use trie, which doesn't do path compression at all, it costs you even more space because each char in the prefix is now a trie node with additional memory overhead. 

> also some logic to validate a metric with valid characters is also needed (like char_to_index in trie.c in this PR)

And I'm sure you can still do that. You can also look into regex if you want to do faster validation.

resyfer:
@bassamadnan You can try looking at `prometheus_client.c` (as already mentioned above) and even [bridge.c](http://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/bridge.c#L965) for an example of the usage of the ART API. You really don't need to know the internals once you've figured out what components the API needs or how the API is used. In other words that's the reason why the API exists :D

bassamadnan:
Fortunately the code was very similliar, so I could just take a refference from existing APIs and port it. I have removed trie and just use existing ART.

> https://db.in.tum.de/~leis/papers/ART.pdf

Thanks for this, was a fun read, still abit complicated i'd say and im not sure if i completely understand it. But the variables and structs do make much more sense now.



> 
> I don't understand the concern here, can you elaborate? `MAX_PREFIX_LEN` is the size of prefix "cache", we don't save the rest of the prefix if it's longer than that. Having the first 10 chars of a prefix serves as a filter for those whose prefixes diverge before the first 10 characters. If you use trie, which doesn't do path compression at all, it costs you even more space because each char in the prefix is now a trie node with additional memory overhead.

I was  confused with  what `MAX_PREFIX_LEN` meant, reading the paper, watching few videos and seeing the code did clarify it.

> And I'm sure you can still do that. You can also look into regex if you want to do faster validation.

Yup,  just added a simple function in `utils`



> @bassamadnan You can try looking at `prometheus_client.c` (as already mentioned above) and even [bridge.c](https://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/bridge.c#L965) for an example of the usage of the ART API. You really don't need to know the internals once you've figured out what components the API needs or how the API is used. In other words that's the reason why the API exists :D

Yes. This was very helpful. Thanks!

PTAL.

jesperpedersen:
Merged.

Issue #252, opened 2025-06-23
Add support for Timescale


bassamadnan:
**Is your feature request related to a problem? Please describe.**

Yes, we need to support multiple versions of [timescale](https://www.tigerdata.com/), now known as TigerData.

Describe the solution you'd like
Queries in the extensions/ with support for new features added across vesions

What is the SQL statement you would like to support
SQL queries must be fitting with utilities that timescale provide and should be relevant to `pgexporter`.

PR #264, opened 2025-07-12
Add timescale metrics

bassamadnan:
Addressing #252 , Added TigerData (timescale) metrics, atleast what is relevant for `pgexporter`.  Here are the extraction scripts, using docker for base extension versions (x.x.0) of every version supported from Postgres 13 onwards. 

I have also tried to test them (some tests are meant to be skipped due to unavailability of the schema). Those results can be found [here](https://github.com/bassamadnan/pg_ext_tracker/tree/478e5978d0090ee0049421e85792c6aa714b539e/timescale_db/timescale_feature_tests). This is the [script](https://github.com/bassamadnan/pg_ext_tracker/blob/478e5978d0090ee0049421e85792c6aa714b539e/timescale_db/run.sh) to run the docker scripts in turn and also contains details of what we focused on and what was extracted accross the versions for checking `diff`

### DIFF
--- a/extensions/timescaledb.yaml
+++ b/extensions/timescaledb.yaml
@@ -1,0 +1,332 @@
+#
+# Copyright (C) 2025 The pgexporter community
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list
+# of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this
+# list of conditions and the following disclaimer in the documentation and/or other
+# materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

jesperpedersen:
@bassamadnan I think you should work on a `contrib/ext_scripts` where you can put your scripts that generates the YAML files

bassamadnan:
Hi @jesperpedersen , i manually generate those YAML files with the help of AI. I do have scripts for testing those (accross different versions) and extracting the differences in the views of these versions, let me know if that'd be a good addition to the `contrib` directory.

jesperpedersen:
@bassamadnan You can always submit a pull request, and we can look at it. It must be something that people can use years from now

bassamadnan:
> @bassamadnan You can always submit a pull request, and we can look at it. It must be something that people can use years from now

Highly agree, will do that.

Need to organize things abit, currently the repository i link is sort of a playground for me. Will submit PRs after things are sorted on that end.

Issue #263, opened 2025-07-12
Custom metrics: Memory leak


jesperpedersen:
```
==860552==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 2 object(s) allocated from:
    #0 0x556b6fd587d8 in malloc (build/src/pgexporter+0xa47d8) (BuildId: b49942e0728a3bfef18a95b547fec3207e917a17)
    #1 0x7ff7e829f3d7 in custom_metrics src/libpgexporter/prometheus.c:1799:28
    #2 0x7ff7e829606c in metrics_page src/libpgexporter/prometheus.c:694:10
    #3 0x7ff7e8294250 in pgexporter_prometheus src/libpgexporter/prometheus.c:243:7
    #4 0x556b6fdae732 in accept_metrics_cb src/main.c:1549:7
    #5 0x7ff7e8ddf5d2 in ev_invoke_pending libev/ev.c:3770:11

SUMMARY: AddressSanitizer: 80 byte(s) leaked in 2 allocation(s).
```

jesperpedersen:
```
2025-07-12 08:08:14 FATAL queries.c:1092 Can not connect to database: test.
2025-07-12 08:08:14 WARN  prometheus.c:1812 Error connecting to server: -475393556, database: 0
```

That is an `INFO` - and just one log statement

jesperpedersen:
@resyfer Let see if we can get this fixed soon

PR #266, opened 2025-07-12
[#263] Custom metrics: Memory leak

resyfer:


jesperpedersen:
@resyfer That part is good -- but we need to change the logging level

jesperpedersen:
@resyfer See what works best - one `INFO`, and you can put the rest at `DEBUG` or `TRACE`

bassamadnan:
I think we need to swap the order of server, database inside the log

resyfer:
I'll add some more logs and update this and push again.

@bassamadnan Yep I had mentioned this issue in the previous PR discussion. Also please use `./uncrustify` for your files before sending a PR (use a pre-commit hook), some are not formatted correctly and it appears as unstaged changes for me.

resyfer:
@jesperpedersen PTAL.

Removed the db counter for when querying on all dbs is disabled (so only `postgres` is queried) as due to the implementation details it would always show `n/n` where `n` is the number of dbs

jesperpedersen:
@resyfer You need to change/remove

```
2025-07-13 08:18:15 FATAL queries.c:1092 Can not connect to database: test.
```

and resolve the server name, so

```
2025-07-13 08:18:15 INFO  prometheus.c:1821 Error connecting to server: 0, database: test
```

to

```
2025-07-13 08:18:15 INFO  prometheus.c:1821 Error connecting to server: primary, database: test
```

jesperpedersen:
@resyfer You forgot to change the server index to the server name

resyfer:
@jesperpedersen oops. Fixed it. Thanks for noticing it.

jesperpedersen:
Merged.

Issue #265, opened 2025-07-12
Add config for extensions


bassamadnan:
We need to have some syntax in place to allow users to control the metrics they would want to load. Across different versions, and different databases (primary, replicas etc) with some logic for handling rules (server > pgexporter > default).

bassamadnan:
This would require changes in `config` and `extension` mostly

PR #269, opened 2025-07-20
Add extensions in config

bassamadnan:
Addressing #265 , added documentation and ran uncrustify for previously missed changes.

### DIFF
--- a/src/include/pgexporter.h
+++ b/src/include/pgexporter.h
@@ -73,7 +73,7 @@ extern "C" {
 #define NUMBER_OF_ENDPOINTS    32
 #define NUMBER_OF_EXTENSIONS   64
 #define NUMBER_OF_DATABASES    64
-#define NUMBER_OF_METRIC_NAMES 256
+#define NUMBER_OF_METRIC_NAMES 1024
 
 #define STATE_FREE        0
 #define STATE_IN_USE      1
@@ -98,6 +98,7 @@ extern "C" {
 
 #define MAX_QUERY_LENGTH      2048
 #define MAX_COLLECTOR_LENGTH  1024
+#define MAX_EXTENSIONS_CONFIG_LENGTH  2048
 
 #define LABEL_TYPE      0
 #define COUNTER_TYPE    1
@@ -270,6 +271,8 @@ struct server
    char tls_ca_file[MAX_PATH];                                  /**< TLS CA certificate path */
    char databases[NUMBER_OF_EXTENSIONS][DB_NAME_LENGTH];        /**< Databases in the server */
    struct extension_info extensions[NUMBER_OF_EXTENSIONS];      /**< The extensions */
+   char extensions_config[MAX_EXTENSIONS_CONFIG_LENGTH];        /**< Server-specific extensions configuration */
+
 } __attribute__ ((aligned (64)));
 
 /** @struct user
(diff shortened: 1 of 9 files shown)


src/libpgexporter/extension.c
```diff
@@ -380,4 +380,66 @@ pgexporter_load_single_extension_yaml(char* extensions_path, char* extension_nam
       fclose(file);
    }
    return 1;
+}
+
+bool
+pgexporter_extension_in_list(char* extension_name, char* extensions_list)
```
resyfer: I think you can convert this into a static function and remove the prefix?

src/libpgexporter/queries.c
```diff
@@ -963,7 +963,8 @@ pgexporter_detect_extensions(int server)
       }
       else
       {
-         config->servers[server].extensions[extension_idx].enabled = true;
+         config->servers[server].extensions[extension_idx].enabled =
+            pgexporter_extension_is_enabled(config, server, config->servers[server].extensions[extension_idx].name);
```
resyfer, line 967: You're not using the returned value anywhere? And these functions don't seem to have any side-effects as well, so this function call doesn't look like it's doing anything helpful.

Also, weird indentation. Make sure to use `uncrustify` if not done already. Just set it up as a git hook ([pre-commit](https://pre-commit.com/)).

src/libpgexporter/extension.c
resyfer: There's a style of writing that a lot of C codebases use, where if a string they take in as a parameter remains unchanged, they mark it as `const char *` (while `char * const` being for pointers that remain unchanged, and their combinations).

The compiler has no problem in either, and a good one can figure out the implicit const, but it's for the developer's benefit.

resyfer (changes requested):
Some debug statements would be nice as well.

resyfer:
Other than these changes, it lgtm.

bassamadnan:
> Some debug statements would be nice as well.

I agree, will add those

src/libpgexporter/extension.c
bassamadnan: ah yes, will do that.

src/include/pgexporter.h
```diff
@@ -379,6 +382,9 @@ struct configuration
    char log_line_prefix[MISC_LENGTH]; /**< The logging prefix */
    atomic_schar log_lock;             /**< The logging lock */
 
+   char global_extensions[MISC_LENGTH];           /**< Global extensions configuration */
```
jesperpedersen: is `MISC_LENGTH` enough ?

src/include/pgexporter.h
```diff
@@ -379,6 +382,9 @@ struct configuration
...
+   bool global_extensions_set;                    /**< Whether global extensions config was specified */
```
jesperpedersen: `strlen(global_extensions) == 0` ?

src/include/pgexporter.h
bassamadnan: Nope, that will need to change. (PS- `NUMBER_OF_METRIC_NAMES` wasn't enough so changed that as well)

src/include/pgexporter.h
bassamadnan: yes that would be enough too, will do that

src/include/pgexporter.h
jesperpedersen: Just call them both `extensions`

jesperpedersen:
Merged.

jesperpedersen:
@bassamadnan We need a tutorial for extensions at some point

bassamadnan:
Yup, agreed @jesperpedersen Will add that

Issue #284, opened 2025-08-02
Add support for pg_buffercache


bassamadnan:
`pg_buffercache` has alot of interesting use cases for `pgexporter` . Having them would be great.

PR #285, opened 2025-08-02
Add metrics for pg_buffercache

bassamadnan:
Addressing #284. Basic functionality of `pg_buffercache` has worked since version `1.3` (default in pg13-pg15). This includes buffer utilization, top cached relations, cache hit (this one might be abit redundant with one of the core metrics, but I kept it)

Version 1.4 introduced `pg_buffercache_summary`, `pg_buffercache_usage_counts`for aggregated stats and usage distribution. The final version 1.5 (default in pg17) introduced an evict function, but its not going to run with `pg_monitor` anyway so I have excluded it. Hope this looks fine. [Here](https://github.com/bassamadnan/pg_ext_tracker/tree/9ed1303f203d95af93efacc6431b962608040214/pg_buffercache) are the patch differences and test cases.

### DIFF
--- a/extensions/pg_buffercache.yaml
+++ b/extensions/pg_buffercache.yaml
@@ -1,0 +1,219 @@
+#
+# Copyright (C) 2025 The pgexporter community
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Merged.

Issue #288, opened 2025-08-10
Metric_name warning


jesperpedersen:
```
/pgexporter/src/libpgexporter/json_configuration.c:950:16: warning: 'strncpy' output may be truncated copying 127 bytes from a string of length 127 [-Wstringop-truncation]
  950 |                strncpy(config->metric_names[config->number_of_metric_names],
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  951 |                        final_metric_name,
      |                        ~~~~~~~~~~~~~~~~~~
  952 |                        MISC_LENGTH - 1);
      |                        ~~~~~~~~~~~~~~~~

In function 'semantics_yaml',
    inlined from 'pgexporter_read_yaml_from_file_pointer' at /pgexporter/src/libpgexporter/yaml_configuration.c:497:11:
/pgexporter/src/libpgexporter/yaml_configuration.c:1531:16: warning: 'strncpy' output may be truncated copying 127 bytes from a string of length 127 [-Wstringop-truncation]
 1531 |                strncpy(config->metric_names[config->number_of_metric_names],
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1532 |                        final_metric_name,
      |                        ~~~~~~~~~~~~~~~~~~
 1533 |                        MISC_LENGTH - 1);
      |                        ~~~~~~~~~~~~~~~~

In function 'semantics_extension_yaml',
    inlined from 'pgexporter_read_yaml_from_file_pointer' at /pgexporter/src/libpgexporter/yaml_configuration.c:489:11:
/pgexporter/src/libpgexporter/yaml_configuration.c:1726:16: warning: 'strncpy' output may be truncated copying 127 bytes from a string of length 127 [-Wstringop-truncation]
 1726 |                strncpy(config->metric_names[config->number_of_metric_names],
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1727 |                        final_metric_name,
      |                        ~~~~~~~~~~~~~~~~~~
 1728 |                        MISC_LENGTH - 1);
      |                        ~~~~~~~~~~~~~~~
~```

Userfrom1995:
@jesperpedersen May I take this issue?

[...]

PR #289, opened 2025-08-10
[#288] Metric_name warning

Userfrom1995:


jesperpedersen:
@Userfrom1995 This doesn't look right. The problem is the last character in the array since they are equal size. So, we need to have a temporary string (pgexporter_append) and then `memcpy` `MISC_LENGTH-1` into the "value". Looks like there are multiple issues like that

jesperpedersen:
@Userfrom1995 We need the "value" to be `\0` terminated

Userfrom1995:
> @Userfrom1995 We need the "value" to be `\0` terminated

Isn't snprintf make sure It is always null terminated?

jesperpedersen:
@Userfrom1995 Ok, it seems that only Windows is a problem

jesperpedersen:
Merged.

Userfrom1995:
> @Userfrom1995 This doesn't look right. The problem is the last character in the array since they are equal size. So, we need to have a temporary string (pgexporter_append) and then `memcpy` `MISC_LENGTH-1` into the "value". Looks like there are multiple issues like that
Yes, there are multiple places in the codebase that follow the same pattern with strncpy. I can do a separate PR to remove them.

jesperpedersen:
@Userfrom1995 Perfect, just use the same issue

PR #296, opened 2025-08-11
[#288] Replace strncpy with snprintf to ensure null termination.

Userfrom1995:
Replace all strncpy calls with snprintf to eliminate -Wstringop-truncation warnings and ensure strings are always properly null-terminated.


jesperpedersen:
Merged.

Issue #302, opened 2025-08-14
Introduce PROMETHEUS_LENGTH


bassamadnan:
Alot of places we use `MISC_LENGTH` which is relatively small (`128`) sometimes we could require a larger value for storing metric attributes. We can define a new `PROMETHEUS_LENGTH` set to `256` for now which should be used by all prometheus related variables consistently.

bassamadnan:
working on this currently, please assign @jesperpedersen

PR #305, opened 2025-08-15
Use PROMETHEUS_LENGTH

bassamadnan:
Addressing #302 . This variable is used for prometheus names, tags, description. Did `git grep` on those and edited wherever it made sense to.


jesperpedersen:
Merged.

Issue #294, opened 2025-08-10
Bloat estimation


jesperpedersen:
```
SELECT
    schemaname,
    tblname,
    pg_size_pretty(real_size) AS table_size,
    pg_size_pretty(extra_size) AS bloat_size,
    ROUND(extra_ratio, 2) AS bloat_ratio_pct
FROM (
    SELECT
        schemaname,
        tblname,
        bs*tblpages AS real_size,
        (tblpages-est_tblpages)*bs AS extra_size,
        CASE WHEN tblpages > 0
            THEN 100 * (tblpages-est_tblpages)/tblpages::float
            ELSE 0
        END AS extra_ratio
    FROM (
        SELECT
            n.nspname AS schemaname,
            c.relname AS tblname,
            c.relpages AS tblpages,
            CEIL(c.reltuples/((current_setting('block_size')::integer-24)/100)) AS est_tblpages,
            current_setting('block_size')::integer AS bs
        FROM pg_class c
        JOIN pg_namespace n ON n.oid = c.relnamespace
        WHERE c.relkind = 'r'
    ) AS bloat_calc
) AS bloat_summary
WHERE extra_ratio > 20
ORDER BY extra_size DESC;
```

bassamadnan:
why dont we stick this in with the existing queries? Do we need to check versions first?

jesperpedersen:
@bassamadnan What do you mean ? It should go into `internal.h` and friends

bassamadnan:
> [@bassamadnan](https://github.com/bassamadnan) What do you mean ? It should go into `internal.h` and friends

Yup, thats what i meant... Can i take this up?. Will run this query in different postgres versions before determining which version it works for/

jesperpedersen:
@bassamadnan Yes, PostgreSQL 13 -> 17

bassamadnan:
Works fine for all versions 13->17. Example output
```sql
postgres=> SELECT
    schemaname,
    tblname,
    pg_size_pretty(real_size::bigint) AS table_size,
    pg_size_pretty(extra_size::bigint) AS bloat_size,
    ROUND(extra_ratio::numeric, 2) AS bloat_ratio_pct,
    tblpages,
    est_tblpages,
    bs as block_size
FROM (
    SELECT
        schemaname,
        tblname,
        bs*tblpages AS real_size,
        (tblpages-est_tblpages)*bs AS extra_size,
        CASE WHEN tblpages > 0
[...]

PR #306, opened 2025-08-16
Add bloat estimation metric

bassamadnan:
Addressing #294, this query works for every version from pg13->17.


jesperpedersen:
Merged.

Issue #295, opened 2025-08-11
Add verification checks 


bassamadnan:
Based on the discussion we had at #287. We think adding some checks while loading the metrics will help developers with how they name the metric files and the related `extension` field.

bassamadnan:
Documentation will be handled on a new issue: #309. 

Changed the description of this one to handle the same.

PR #310, opened 2025-08-16
[#295] Add verification checks

bassamadnan:
1. Check for same name and file path for extension while loading
2. Handle alphanumerical version abit more robustly

PTAL - @jesperpedersen @resyfer

### DIFF
--- a/src/libpgexporter/extension.c
+++ b/src/libpgexporter/extension.c
@@ -123,6 +123,7 @@ pgexporter_parse_extension_version(char* version_str, struct version* version)
    char* str_copy = NULL;
    char* token = NULL;
    char* saveptr = NULL;
+   char* dash_pos = NULL;
    int part = 0;
 
    if (!version_str || !version)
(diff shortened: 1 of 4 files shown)


src/libpgexporter/extension.c
```diff
@@ -142,42 +143,66 @@ pgexporter_parse_extension_version(char* version_str, struct version* version)
...
       goto error;
    }
 
+   free(str_copy);
    return 0;
 
 error:
-   free(str_copy);
-
+   if (str_copy)
```
jesperpedersen: You don't need the guard - after each `free()` set the variable to `NULL`

jesperpedersen (changes requested):

jesperpedersen:
@bassamadnan And, JSON ?

src/libpgexporter/extension.c
bassamadnan: Right, will make this change

bassamadnan:
> @bassamadnan And, JSON ?

We only deal with extensions in YAML. We dont have any logic for parsing extension files in JSON since we decided to keep all files in `extension` as `.yaml`

jesperpedersen:
@bassamadnan Ok, we can park JSON for later

resyfer (approved):
LGTM, though I don't get the need for `(int) value` typecast.

bassamadnan:
> LGTM, though I don't get the need for `(int) value` typecast.

Thanks! [strtol](https://cplusplus.com/reference/cstdlib/strtol/) returns long value, so just wanted to suppress warnings.

resyfer:
@bassamadnan Hmmm, alright. If the explicit typecast is needed instead of implicit for suppressing warnings, then that's fine by me.

jesperpedersen:
Merged.

Issue #311, opened 2025-08-18
Missing tag in contrib/yaml


bassamadnan:
Some of the files are missing tags (`postgres-13.yaml:893` and 14 and 15 too). Need to add them back

PR #312, opened 2025-08-18
[#311] Missing tag in contrib/yaml

bassamadnan:
Missing tag on these metrics


jesperpedersen:
Merged.

Issue #323, opened 2025-09-03
Add support for pgexporter_ext via extensions


bassamadnan:
Currently the pgexporter_ext quries reside in `queries.c` where we also have a few metrics exposed to prometheus (if the ext is enabled, its version etc). 

Ideally now that we support extensions,  we should move it to the `extensions/` directory along with other extensions and edit the codebase which makes these calls.

PR #314, opened 2025-08-18
[#313] Show database in metrics

bassamadnan:

PR #243, opened 2025-06-14
[#239] Pull metrics from other databases

resyfer:
Allows execution of a query on all databases on a cluster.

Configuration addition:

```yaml
- metrics:
  - queries:
    // ...
    database: all
```

`database: all` executes on all databases, while any other value or not mentioning the key at all executes only on `postgres` database.

NOTE: Custom metrics by users may send duplicate metrics if all databases include a common set of entries for the query, so it's up to the user to ensure no duplication.

PS: `database: *` does not work on YAML, and hence it needed to be `all`.

PR #324, opened 2025-09-03
[#323] Add support for pgexporter_ext via extensions

bassamadnan:


contrib/prometheus_scrape/extra.info
```diff
@@ -3259,4 +3259,185 @@ pgexporter_vector_vector_index_distribution_index_count
 * server: The configured name/identifier for the PostgreSQL server.
 * database: The database being monitored.
 * index_type: The access method of the vector index.
-Version: 0.8.0
\ No newline at end of file
+Version: 0.8.0
+
+pgexporter_pgexporter_ext_extension_version_version
```
jesperpedersen: Where is this coming from ?

src/libpgexporter/queries.c
```diff
@@ -165,107 +165,13 @@ pgexporter_close_connections(void)
    }
 }
 
-int
-pgexporter_query_get_functions(int server, struct query** query)
-{
-   char* d = NULL;
-   int ret;
-
-   d = pgexporter_append(d, "SELECT * FROM pgexporter_get_functions();");
```
jesperpedersen, line 174: Can we remove this in the extension then ?

bassamadnan: yes i suppose we can,  though it did make my life abit easier having a neat table which tells me everything pgexporter_ext has to offer

contrib/prometheus_scrape/extra.info
bassamadnan: new queries from the YAML

jesperpedersen: Yes, but `_version_version` ?

src/libpgexporter/queries.c
jesperpedersen: ok, we can keep it

contrib/prometheus_scrape/extra.info
bassamadnan:
```yaml
queries:
      - query: SELECT pgexporter_version_ext() as version;
        version: "0.3.0"
        columns:
          - name: version
            type: label
            description: pgexporter extension version
``` 

appended from the column here, should i change it to something else? We can completely remove it as the core reports extension version already

jesperpedersen: Just remove for now

bassamadnan: Done, for refference - these are the metrics in on `/metrics` : https://stagb.in/7d6gpbm

jesperpedersen: There are empty database all over ... `database=""` those need to go away

bassamadnan: oh right, I forgot to select database,here are the updated metrics: https://stagb.in/9ntfi2h

resyfer: Wait, wdym "I forgot to select database"? Shouldn't the `postgres` db be the default here?

jesperpedersen:
Merged.

contrib/prometheus_scrape/extra.info
bassamadnan: Ever since [this](https://github.com/pgexporter/pgexporter/pull/243) patch dont we specify the database? (I noticed it abit later, but it does not default to `postgres`) After I noticed [this](https://github.com/pgexporter/pgexporter/pull/314) we applied this patch

jesperpedersen: if it is empty then we should assume "postgres" - the field must always be filed out in metrics where it appears

bassamadnan: Oh, I was under the assumption we have to select it? For eg- the core metrics have this:
```
#HELP pgexporter_pg_replication_slots_active Is the replication active
#TYPE pgexporter_pg_replication_slots_active gauge
pgexporter_pg_replication_slots_active{server="primary", slot_name="repl", slot_type="physical", database=""} 0

#HELP pgexporter_pg_replication_slots_temporary Is the replication temporary
#TYPE pgexporter_pg_replication_slots_temporary gauge
pgexporter_pg_replication_slots_temporary{server="primary", slot_name="repl", slot_type="physical", database=""} 0
```

jesperpedersen:
@bassamadnan Same problem then - all attributes must have a value

bassamadnan:
> @bassamadnan Same problem then - all attributes must have a value

Okay, @resyfer I suppose we change the query for the `replication_slots`? or did you have something else in mind

resyfer:
@bassamadnan It should be handled internally. Just like the case with N/A, all the attributes need a default value...the database key having the default postgres value. I should say N/A would be fine for others.

bassamadnan:
But then why isnt this metric being handeled that way then? Its one of the core metrics so the same logic should apply

jesperpedersen:
The problem is that we don't have a data type for each attribute; in certain cases we can guess, but we likely have to fallback to "n/a" if we don't know.

All metrics should pass-through a "safe_value" sort of function before being printed. This function could look at the previous entry to see if it can guess the value in case there isn't one in the current entry.

Random thoughts

PR #356, opened 2025-11-02
[#355] Add metrics for pgstattuple

bassamadnan:
I have tried to use `approx` views wherever possible since thats better for large tables. A stable extension, simple addition.


jesperpedersen:
Merged.

PR #360, opened 2025-12-07
[#359] Add pre-flight validation of the configuration file before attempting

saraanbih:
## Problem

The daemon may fail while parsing the main configuration file and emit ambiguous parsing errors when the root cause is a filesystem issue (missing file, a directory path, insufficient permissions, or a binary file).
These cases produce unclear operator feedback (including poor systemd STATUS messages) and make diagnosis harder.
Proposed Solution
## Solution

    checks the path exists,
    verifies it is a regular file,
    ensures it is readable,
    ensures the file is not binary (no NUL bytes).

Fixes [#359]

### DIFF
--- a/src/libpgexporter/configuration.c
+++ b/src/libpgexporter/configuration.c
@@ -92,6 +92,7 @@ static bool is_empty_string(char* s);
 
 static void add_configuration_response(struct json* res);
 static void add_servers_configuration_response(struct json* res);
+static bool pgexporter_is_binary_file(const char* path);
 
 /**
  *
@@ -146,6 +147,40 @@ pgexporter_init_configuration(void* shm)
    {
       config->prometheus[i].sort_type = SORT_NAME;
       config->prometheus[i].server_query_type = SERVER_QUERY_BOTH;
+   }
+
+   return 0;
+}
+
+/*
+ *
+ */
+int
+pgexporter_validate_config_file(char* path)
+{
+   if (path == NULL)
+   {
+      return EINVAL;
+   }
+
+   if (!pgexporter_exists(path))
+   {
+      return ENOENT;
+   }
+
+   if (!pgexporter_is_file(path))
+   {
+      return ENOENT;
+   }
+
+   if (access(path, R_OK) != 0)
+   {
+      return EACCES;
+   }
+
+   if (pgexporter_is_binary_file(path))
+   {
+      return EINVAL;
    }
 
    return 0;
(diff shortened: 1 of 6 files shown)


saraanbih:
Hi Mr. @jesperpedersen, I hope you’re doing well. I’ve made the changes you requested in pgexporter. Could you please review them?

src/libpgexporter/configuration.c
```diff
@@ -151,6 +152,35 @@ pgexporter_init_configuration(void* shm)
...
+pgexporter_validate_config_file(char* path)
+{
+   if (path == NULL || !pgexporter_exists(path))
+   {
+      return 1;
+   }
+
+   if (!pgexporter_is_file(path))
+   {
+      return 2;
```
resyfer: Usually when you return values like this, you use enums or macros containing these values (in C it's usually the latter). This way, the user of `pgexporter_validate_config_file` doesn't have to look into the implementation to guess what each error value means.

src/libpgexporter/configuration.c
```diff
@@ -3620,3 +3650,39 @@ is_empty_string(char* s)
...
+   if (fp == NULL)
+   {
+      goto error;
+   }
+
+   while ((bytes = fread(buffer, 1, sizeof(buffer), fp)) > 0)
+   {
+      for (size_t i = 0; i < bytes; i++)
+      {
+         if (buffer[i] == '\0')
```
resyfer: Another check (along with `\0`) for `<= 0x7f` should be there (<= means binary, else it's not).

`\0` is not always a great indicator alone for binary data.

src/libpgexporter/configuration.c
resyfer: It's also a good practice to use err codes already baked into standard C (and often the OS) instead of custom error codes. Take a look at `man 3 errno` (either in your terminal or [here](https://man7.org/linux/man-pages/man3/errno.3.html)). They're also often returned as negative values (like `return -ENOENT;`).

jesperpedersen:
@saraanbih Hi Sara, @resyfer will take care of this one - apply his comments to the pgagroal one as well, then send a delta pull request to pgmoneta under the existing issue number. Thanks !

saraanbih:
@resyfer I have made the changes, Can you review them please?

jesperpedersen:
@saraanbih Lets wait until pgagroal and pgmoneta are green, and then sync this pull request

jesperpedersen:
Merged.

PR #362, opened 2025-12-21
[#363]Add clang-format configuration and script for code formatting

TutTrue:
This PR replaces Uncrustify with ClangFormat as the C code formatter used in this repository while keeping formatting behavior consistent with the existing style.


jesperpedersen:
@TutTrue Remember to create the issue for this and then reference that in the commit message. Try to the first sentence of the commit message short. Put the rest in the body

TutTrue:
Preprocessing /workspace/src/include/internal.h...
Parsing file /workspace/src/include/internal.h...
input buffer overflow, can't enlarge buffer because scanner uses REJECT
    lexical analyzer: ./src/scanner.l (for: /workspace/src/include/internal.h)

Exiting...
make[3]: *** [doc/CMakeFiles/api.dir/build.make:71: api] Error 2
make[2]: *** [CMakeFiles/Makefile2:272: doc/CMakeFiles/api.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:279: doc/CMakeFiles/api.dir/rule] Error 2
make: *** [Makefile:241: api] Error 2

✗ FAILURE: Documentation build failed
Check the error messages above

ci fails because of an overflow, how can i fix this? do i just exclude the internal.h from the Cmake?

jesperpedersen:
> input buffer overflow, can't enlarge buffer because scanner uses REJECT

Can't you change that ?

Otherwise, yes, exclude internal.h which isn't ideal

TutTrue:
> > input buffer overflow, can't enlarge buffer because scanner uses REJECT
> 
> Can't you change that ?
> 
> Otherwise, yes, exclude internal.h which isn't ideal

i didn't get any way rather than excluding the file but it adds no value anyway so i think it is just fine to exclude it

jesperpedersen:
@TutTrue Ok, force push that

PR #366, opened 2026-01-04
[#250] Added web console ui

### LINKED ISSUE #250: Simple web management console
Create a simple web management console that support the management protocol, and dynamically create a UI based on the Prometheus metrics.

"External" frameworks, like https://dygraphs.com/, can be used - but users can download them themselves such that we don't have to bundle them.

Some of the HTTP functionality can like be shared with the Prometheus component, so http.h|.c.

If this becomes a big and complex project then we can move its management to a separate project.

sommye-ctr:
Resolves #250 

## Changes
- Added `console.c` and `console.h` implementing a standalone console layer
- Includes basic management info (status, version, per-server status)
- Includes prometheus metrics, fetched from `prometheus_client.c` 
- Dynamically groups metrics into categories using global prefix analysis:
    - all metric prefixes are counted and candidate categories are considered using a score (avoid overly specific and favours widely shared) to avoid flooding of categories
- Added a JSON endpoint exposing the same console data model for future UI or tooling use

## Extensibility
- The console is a pure data-presentation layer and contains no HTTP server logic
- The branding needs to be passed from wherever the init function is called

## Screenshots
image
image


jesperpedersen:
@sommye-ctr Remember, https://github.com/pgexporter/pgexporter/blob/main/doc/DEVELOPERS.md#authors

sommye-ctr:
@jesperpedersen Done

jesperpedersen:
https://github.com/pgexporter/pgexporter/blob/main/AUTHORS

sommye-ctr:
@jesperpedersen Done

src/libpgexporter/prometheus.c
```diff
@@ -56,7 +57,9 @@
 #define PAGE_UNKNOWN                0
 #define PAGE_HOME                   1
 #define PAGE_METRICS                2
-#define BAD_REQUEST                 3
+#define PAGE_CONSOLE                3
```
jesperpedersen: It need to have a dedicated port - so separate from prometheus.c

src/include/console.h
```diff
@@ -0,0 +1,183 @@
+/*
```
jesperpedersen, line 1: Follow the same path as prometheus.h

sommye-ctr:
@jesperpedersen Made its separate port, which can be specified in the configuration similar to metrics port.
Now the console handles serving the pages similar to how prometheus is doing.

resyfer:
I've taken a look at this. It's a fairly big patch but a good one. On the surface it looks good. Need to test it out a bit more over the next couple of days to catch stray bugs, memory or logical, on my end and then it should be good to go.

FYI @sommye-ctr, all the non-static functions in header files (ie. the ones that will be accessed outside of the corresponding `.c` file) are prefixed with `pgexporter_` (I saw that you already had them, though I have no idea why you removed the prefix).

src/main.c
```diff
@@ -1661,6 +1736,112 @@ accept_metrics_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+         pgexporter_log_debug("accept: %s (%d)", strerror(errno), watcher->fd);
+      }
+      errno = 0;
+      return;
+   }
+
+   pid = fork();
+   if (pid == -1)
+   {
+      pgexporter_log_error("Console: No fork (%d)", MANAGEMENT_ERROR_METRICS_NOFORK);
```
jesperpedersen: ?

src/main.c
```diff
@@ -1661,6 +1736,112 @@ accept_metrics_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+      goto error;
+   }
+   else if (pid == 0)
+   {
+      ev_loop_fork(loop);
+
+      /* We are leaving the socket descriptor valid such that the client won't reuse it */
+      shutdown_ports();
+      if (strlen(config->metrics_cert_file) > 0 && strlen(config->metrics_key_file) > 0)
```
jesperpedersen: ?

src/main.c
```diff
@@ -1661,6 +1736,112 @@ accept_metrics_cb(struct ev_loop* loop, struct ev_io* watcher, int revents)
...
+      {
+         if (pgexporter_create_ssl_ctx(false, &ctx))
+         {
+            pgexporter_log_error("Could not create console SSL context");
+            return;
+         }
+
+         if (pgexporter_create_ssl_server(ctx, config->metrics_key_file, config->metrics_cert_file, config->metrics_ca_file, client_fd, &client_ssl))
```
jesperpedersen: ?

sommye-ctr:
@jesperpedersen Ahh, I had reused most of that code from the existing metrics/management segment, those things got overlooked, sorry for that...
Currently, the console port is taken by the configuration file and it is running without ssl context.

sommye-ctr:
> FYI @sommye-ctr, all the non-static functions in header files (ie. the ones that will be accessed outside of the corresponding `.c` file) are prefixed with `pgexporter_` (I saw that you already had them, though I have no idea why you removed the prefix).

@resyfer I was bulk replacing that for the metric prefix - must have gotten removed then

sommye-ctr:
Any updates on this?

jesperpedersen:
@sommye-ctr @resyfer is busy right now - you are going in the right direction so keep working at it

bassamadnan:
@sommye-ctr Maybe from a UI prespective, we can also look at existing dashboards to have more visualizations. Googling "postgres monitoring dashboards" will give you some existing UI which seem to be a good fit. Some of the ones I could find were, ofcourse existing Grafana templates, Metabase, pgDash etc. Though they are paid solutions, their visuals do provide insights on what sorts of "marks and channels" work the best for different metrics. 

Although I am not completely sure on the scopre of this issue/PR as this may as well deserve its own repo under the pgexporter org.

jesperpedersen:
We shouldn't go too crazy - I like the `console.h|.c` approach - but if we can improve the UX inside those files then it is of course a good thing.

We should support the basic use-cases such that developers and small installations are covered for their needs

sommye-ctr:
We can have the dygraphs functionality as mentioned in the issue - make it such that if it is installed by the user, show some visualisations. As for the base UI/UX (without dygraphs) - I have deliberately kept it minimal. If some changes are requested in that part I could look into it.

jesperpedersen:
Having graphs would be a good thing, but lets wait a bit with that. I'll give this a deep review - hopefully this week

jesperpedersen:
@sommye-ctr You need documentation for this

jesperpedersen:
* "pgexporter Console" -> "Web console"
* "INFO  console.c:1834 pgexporter_console: start" -> "DEBUG  console.c:1834 pgexporter_console: start"
* "1.00" -> "1" when there is no comma value
* Test with a setup with a primary and a replica since values are missing
* "pg_blocked_vacuum_blocked_duration" + "pg_blocked_vacuum_blocking_duration" -> "pg_blocked_vacuum" tab
* Create a "Simple" view (Name, Value) and "Full" view (Name, Gauge, Value, Labels). Simple should be default

jesperpedersen:
Maybe just start by removing "pgexporter_" and then group... lets see how that looks

jesperpedersen:
In Simple we will need a "Server" column that is based on the "server" label

src/include/console.h
```diff
@@ -0,0 +1,191 @@
...
+struct console_status
+{
+   char* status;                   /**< Overall status */
+   char* version;                  /**< pgexporter version */
+   int num_servers;                /**< Number of configured servers */
+   char* last_updated;             /**< ISO timestamp of last update */
+   struct console_server* servers; /**< Array of server information */
+};
+
+/**
```
jesperpedersen: Do we need all of this in .h ? Can't most of it be moved to the .c file ?

jesperpedersen:
@sommye-ctr Maybe instead of having a "Server" column we should start with a mouse-over text box with the remaining information. Then we can figure out how to display it better since some of the labels contain important information

jesperpedersen:
And, have "Name" be a link to a dedicated page where all labels have their own column

jesperpedersen:
Try various combinations of the feedback - it should be easy to find `primary` and `replica` information

jesperpedersen:
See you can do a "Search" box with auto-completion of full metric names (pgexporter_...) such that the DBA can quickly find a specific metric

resyfer:
@jesperpedersen Would a simple HTML dropdown menu be better than the list of metrics written side by side? That frees up the screen real-estate for cases when there are a lot of metrics.

Dropdown menu can have an autocomplete logic. The logic would be like if a person has typed `abc`, it would match the regex `*abc*`.

I think it's best to keep a reliance on JS and HTML as much as possible. Too flashy UIs are both hard to maintain, nightmare for responsiveness on multiple devices, and distracts from the actual data.

I feel the current UI is good enough.

@sommye-ctr Apart from the above, I've some further suggestions. If console's enabled, you should add a link to it on the home page of `pgexporter` (as in, when you visit `localhost:5002`), similar to the `Metrics` page.

Further, I would like to see the labels as a table as well (nested). They are key value pairs separated by a comma delimiter.

src/include/console.h
resyfer: Keep stuff in `.h` files that you mean to use in other files, in which case add a `pgexporter_` prefix to their identifiers.

Stuff that are only used in its corresponding `.c` file should be present in the `.c` file only (like how modules are). In this case try using the `static` keyword wherever applicable (mostly in the function definitions).

jesperpedersen:
@resyfer Yes, strings are a PITA but few humans can read a byte dump array

jesperpedersen:
The goal of the web console is to have a functional console that displays Prometheus data in a nicer way, and it will improve with the GSoC 2026 project. It isn't meant to be a full Grafana implementation, but it should cover most common use-cases

jesperpedersen:
@sommye-ctr Keep going down the single `console.h` and `console.c` path. We don't want a million files for this

resyfer:
@jesperpedersen I meant about doing this conversion step during compilation. It's not required though, but it would be nicer. It can't be used for the HTML part of the web console anyway as the HTML has to be generated using logic.

sommye-ctr:
> * "pg_blocked_vacuum_blocked_duration" + "pg_blocked_vacuum_blocking_duration" -> "pg_blocked_vacuum" tab

This is actually the main problem with the dynamic metrics coming in - this particular thing behaves like an edge case. I decided to create categories which have min of 3 metrics but the `pg_blocked_vacuum` has only 2 and therefore they got into a separate tab altogether. I could play around with the scoring and other constants for the algorithm to get the desired result.

> I also don't get the point of the `x` on the servers.

They were kind of like status for pgexporter - whether those servers are active and connected.

Also, I do think the dropdown menu would be a good choice to handle the various metrics, it will also cover the autocompletion with search functionality. I'll start working on the feedbacks and see what covers most of the basic functionality.

jesperpedersen:
@sommye-ctr Don't go too crazy before pushing your branch, and sharing screenshots

sommye-ctr:
image
image
image


How does this look? As said above, the names can be made clickable and redirect to new page for columns of labels. The servers dropdown acts as a filter to get metrics for that specific server.

jesperpedersen:
@sommye-ctr You can make "Simple" and "Detailed" a drop-down as well. See if you can make the `1.00` -> `1` work as well

sommye-ctr:
image
How's this?

jesperpedersen:
@sommye-ctr Excellent

sommye-ctr:
image
image
The name for metric can be clickable like this which leads to new page and such a table with all labels as columns.

jesperpedersen:
@sommye-ctr Yeah, that looks good

sommye-ctr:
I am a little unsure about the use case of it tho. Like in the detailed view this is already shown so the friction of clicking and directing to a new page just to see a single row of the same data in a 'prettier' format...

Maybe we add a new 'view' which groups the metrics (of a particular category) with same labels and then shows multiple tables with their labels as columns

jesperpedersen:
@sommye-ctr Yes, that is the idea - all the "same" metrics should be in the same detailed view when you press the link

jesperpedersen:
@sommye-ctr Are you ok on this ? We can always improve

sommye-ctr:
@jesperpedersen I was a little busy the previous week. I'll give more updates

sommye-ctr:
> @sommye-ctr Yes, that is the idea - all the "same" metrics should be in the same detailed view when you press the link

Could you elaborate on this? I think I am getting a wrong idea about it

jesperpedersen:
@sommye-ctr I meant if a user press `pg_vacuum` then all `pg_vacuum_XYZ` are shown. But, see how it will work/shown

sommye-ctr:
Maybe the category heading be made clickable, like heading `pg_settings` -> detailed page with all the labels as columns.
image
image

Though, on the other hand there would be a page like `pg_stat` too:
image

jesperpedersen:
@sommye-ctr I think we should start with that :+1:

sommye-ctr:
Currently the server filter is non functional since I am figuring out its multiple select UI.

sommye-ctr:
@jesperpedersen Should we go towards a darker mode ui? or light is fine?

jesperpedersen:
@sommye-ctr If it is easy to make then you can make a toggle icon in top-right corner

sommye-ctr:
Added the dark mode toggle. Also made the server functionality working with multi select dropdown 
image

Also, made most of the functions static and moved them from .h

jesperpedersen:
@sommye-ctr Great ! For `console.c` use the API from `utils.h` - like `pgexporter_append` instead of `strcat`, `pgexporter_snprintf` instead of `snprint` and so on.

That should align the code, and make it easier. Once done, I'll give it a deeper review

sommye-ctr:
I've now used the pgexporter utility. The patch turned out to be a huge one!

jesperpedersen:
@sommye-ctr Lets start with documentation - README.md, doc/CONSOLE.md, doc/manual/en/10-console.md (move existing 1x up by one), and screenshots. Remember the .spec file too.

Think from `* Web console` to a brief user guide with screenshots (light mode)

sommye-ctr:
@jesperpedersen How's this?

jesperpedersen:
@sommye-ctr Overall it looks very good ! Comments,

* What is "Status: Disconnected" ?
* If I have a primary and replica and go to "postgresql" I only see the replica entries (my filter is All)
* `Category: postgresql` -> `postgresql`
* Select "postgresql", switch between "Simple" and "Detailed" `uptime` is the same. Wait, press "postgresql", and the "Category" view shows the correct uptime. So, somewhere the cache isn't updated...
* In dark mode - probably make the links white with the underline

... I'm thinking about the "Simple" view ... but we can look at that later as well.

jesperpedersen:
@sommye-ctr My thinking is how to display "postgresql" in "Simple" with the primary and replica server. Maybe an idea could be to do the same as "Category" and filter out the "endpoint" attribute", leave "Detailed" as is, and remove the "Category" view... That is my idea atm

sommye-ctr:
> @sommye-ctr Overall it looks very good ! Comments,
> 
> * What is "Status: Disconnected" ?
> * If I have a primary and replica and go to "postgresql" I only see the replica entries (my filter is All)
> * `Category: postgresql` -> `postgresql`
> * Select "postgresql", switch between "Simple" and "Detailed" `uptime` is the same. Wait, press "postgresql", and the "Category" view shows the correct uptime. So, somewhere the cache isn't updated...
> * In dark mode - probably make the links white with the underline
> 
> ... I'm thinking about the "Simple" view ... but we can look at that later as well.

hmm, these all seem to be bugs. I'd look into this.
Also is status field desired? I mentioned what it is intended to show in the docs (currently it is buggy) - so is it something which is desired or we can remove it altogether?



> @sommye-ctr My thinking is how to display "postgresql" in "Simple" with the primary and replica server. Maybe an idea could be to do the same as "Category" and filter out the "endpoint" attribute", leave "Detailed" as is, and remove the "Category" view... That is my idea atm

I agree endpoint attribute can be removed. As for the server maybe we have 3 columns in "Simple" view showing the server too (if we look, it doesnt make much sense to expose the server filter and not show the server at all in the simple view). Also the category view, um I am unsure about it too.

jesperpedersen:
@sommye-ctr I think we should try and see how all columns would look like to begin with (except endpoint). Then we could rename "Detailed" to "Advanced". WDYT ?

sommye-ctr:
So "Simple" -> Name | Value | Server
"Advance" -> Name | Type | Value | Server | Labels
and no "Category" view (labels as columns) for now?

jesperpedersen:
@sommye-ctr No, "Simple" as the "Category" is now without the "endpoint" column. "Detailed" renamed to "Advanced" - no other changes. Remove the "Category" view, and therefore the link to it

jesperpedersen:
@sommye-ctr Remember the format

jesperpedersen:
@sommye-ctr There is compile issues too, so rebase

jesperpedersen:
@sommye-ctr `Status: Disconnected` is still there, but importantly only `replica` is shown and nothing from `primary`.

"Simple" is missing the columns, and "Advance" (should be Advanced) is missing its "endpoint"

sommye-ctr:
I was treating the definitions of prometheus metric wrongly and therefore only `replica` ones were showing. Fixed that. Also renamed the status to service and now it shows whether the management service is running or unavailable.

Also Simple view -> Name | Value | Columns
Advanced view -> Name | Type | Value | Labels

jesperpedersen:
@sommye-ctr Can you rebase ? We should probably wait for #416 first

jesperpedersen:
The following is from primary/replica with PostgreSQL 17

```
src/libpgexporter/console.c:1731:20: runtime error: 2.33113e+09 is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/libpgexporter/console.c:1731:20 
```

jesperpedersen:
@sommye-ctr We are getting really close now. We need to refresh the data if the view (Simple/Advanced) or Category is switched. Right now, you have to refresh manually in order to get the updated values

jesperpedersen:
@sommye-ctr An idea could be to have an explicit "Refresh" button/dropdown

```
... | Updated ... | Refresh
```

such a refresh of the data is done when pressed, or by the interval selected

jesperpedersen:
@sommye-ctr ... maybe a default of 5 minute ?

sommye-ctr:
Maybe we have a configuration attribute in .conf file like `console_refresh_duration` (with a default value of say 5 min) plus a refresh button to refresh it manually? WDYT?

jesperpedersen:
@sommye-ctr I don't think we need it in the configuration right now - lets start with an explicit button

sommye-ctr:
So just to be clear, a button "Refresh" which refreshes when clicked + an option to select the interval for auto-refresh (defaulting to 5 min) be shown in the console?

jesperpedersen:
@sommye-ctr It depends on how easy it is to do - either a button that users has to press, or a dropdown that can reload automatically depending on the setting (default to 5m).

So, one "thing" - as simple as possible

sommye-ctr:
I haven't had the time to think too much about this (due to my exams)...but
I feel that the complicated bit would be refreshing ONLY the values. Like the easiest solution would be to reload the entire webpage when clicked (but we would lose the pages state with it). 

On the other hand - a sophisticated way would be to use the json endpoint to get the json result of the updated metrics (in js) and then update the table rows using that.

Maybe we can do is store the "state" in local storage (like the category selected, filters etc..) -> then reload the entire page -> then load those state. 
@jesperpedersen WDYT

jesperpedersen:
@sommye-ctr Don't overcomplicate right now - a full reload right now is ok. Later we can look into partial reloads

jesperpedersen:
@sommye-ctr Great work !

jesperpedersen:
Merged.

jesperpedersen:
@sommye-ctr Please, port to pgmoneta and pgagroal

Issue #369, opened 2026-01-11
Add global PostgreSQL query timeout support for metric collection


sommye-ctr:
**Is your feature request related to a problem? Please describe.**

Currently, pgexporter executes multiple SQL queries during a Prometheus scrape without enforcing any query-level timeout. If a single metric query becomes slow or blocked, it can stall the entire scrape. This causes Prometheus to time out the `/metrics` endpoint and incorrectly mark the PostgreSQL instance as down, even though the database itself is healthy.

**Describe the solution you'd like**

Introduce a configurable global query timeout for all metric SQL execution using PostgreSQL’s `statement_timeout`. The timeout would be set before executing each metric query so that slow or blocked queries are canceled by PostgreSQL, allowing pgexporter to continue collecting and exposing the remaining metrics.

**What is the SQL statement you would like to support**

The feature relies on setting PostgreSQL’s session-level timeout before running metric queries:
`SET statement_timeout = <configured_timeout_in_ms>;`

**Additional information**

The timeout would be configured once by the user, in the main pgexporter configuration

sommye-ctr:
@jesperpedersen Any thoughts on this?

jesperpedersen:
@sommye-ctr Yes, this feature is valid

sksingh2005:
@sommye-ctr sir are you working on it ??

sommye-ctr:
@sksingh2005 Yes I am working on this

PR #370, opened 2026-01-20
[#369] Added global postgresql query timeout

sommye-ctr:
### Changes
- Introduced a new `metrics_query_timeout`,  allowing users to set a global timeout (in milliseconds) for all metric SQL queries.
- Implemented logic in the connection setup and database switch routines to automatically set the PostgreSQL `statement_timeout` for all metric connections, enforcing the configured timeout.
- Added minimum value check of 50ms to prevent too low timeouts.
- Added a new function `pgexporter_execute_command` to execute the queries that dont return dataset.

### Testing
- Tested this functionality by setting a too low timeout (bypassing the 50ms constraint) and the `settings` information was skipped due to the timeout
- Tested by adding a custom metrics which sleeps for 3 seconds and the timeout caused it to skip that request.

### DIFF
--- a/src/libpgexporter/queries.c
+++ b/src/libpgexporter/queries.c
@@ -52,6 +52,7 @@ static int process_server_parameters(int server, struct deque* server_parameters
 static int pgexporter_detect_databases(int server);
 static int pgexporter_detect_extensions(int server);
 static int pgexporter_connect_db(int server, char* database);
+static void pgexporter_apply_metrics_timeout(int server);
 
 int
 pgexporter_check_pg_monitor_role(int server)
@@ -176,6 +177,8 @@ pgexporter_open_connections(void)
 
             pgexporter_detect_databases(server);
             pgexporter_detect_extensions(server);
+
+            pgexporter_apply_metrics_timeout(server);
          }
          else
          {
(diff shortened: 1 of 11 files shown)


src/libpgexporter/queries.c
```diff
@@ -176,6 +176,33 @@ pgexporter_open_connections(void)
 
             pgexporter_detect_databases(server);
             pgexporter_detect_extensions(server);
+
+            /* Set query timeout for metrics if configured */
+            if (config->metrics_query_timeout > 0)
+            {
+               if (config->metrics_query_timeout < 50)
```
jesperpedersen: You can move this to configuration.c and validate_configuration

src/libpgexporter/queries.c
```diff
@@ -1066,6 +1180,32 @@ pgexporter_switch_db(int server, char* database)
       goto error;
    }
 
+   /* Set query timeout for metrics if configured on the new connection */
+   if (config->metrics_query_timeout > 0)
+   {
+      if (config->metrics_query_timeout < 50)
```
jesperpedersen: Remove this

sommye-ctr: Now that I see, maybe this should go in the connect_db function instead of the switch_db for the custom metrics timeout.

jesperpedersen: yeah, you can try that

jesperpedersen:
@sommye-ctr You forgot the AUTHORs stuff

jesperpedersen:
Merged.

PR #376, opened 2026-01-27
Fix missing libatomic in RPM Dockerfile

cracksalad:
While using the RPM Dockerfile to build a RPM, I came across the following error, which is easy to fix:

```
CMake Error at CMakeLists.txt:180 (message):
  libatomic needed


-- Configuring incomplete, errors occurred!
```


jesperpedersen:
Merged.

Issue #377, opened 2026-01-29
pgexporter_pg_statio_* metrics


benjunmun:
I believe that the current pg_statio related metrics are not running a useful query. The [current query](https://github.com/pgexporter/pgexporter/blob/5cc503380d6827f5981b3f1ef17bf7ecd195a63d/src/include/internal.h#L491) involves a COUNT on the columns of pg_statio_all_tables - this just counts the rows in this table that have a non-null value. 

I believe the minimum change to get meaningful data would be to SUM across the columns so that the data is at least meaningful. A more invasive change would be to add labels on schema and relname to avoid the sum entirely.

The underlying postgres pg_statio_all_tables definition is documented [here](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STATIO-ALL-TABLES-VIEW).

I'll prepare a pull request to switch COUNT to SUM in case that is useful for the project.

jesperpedersen:
@benjunmun To me it sounds like the `relname` direction is the correct way

benjunmun:
Fair enough, I think it's more useful that way as well. Additionally, I believe this query should be run on all databases since it relates to the actual tables and not just the system as a whole. I'll send a new PR with these changes.

jesperpedersen:
@benjunmun Thanks !

PR #378, opened 2026-01-29
Fix pg_statio metrics

benjunmun:
Previous query was effectively counting tables, instead of combining the data across each table. This change SUMs the data from pg_statio_all_tables and pg_statio_all_sequences, instead of counting rows from those tables.

This is a minimal fix for issue #377


benjunmun:
As requested, submitting a fix for this in a different way.

PR #379, opened 2026-01-30
Make pg_statio metrics more useful

benjunmun:
* Fix queries to report actual stats by relname instead of counting relations.
 * Run statio queries on all databases since they relate to the actual tables and not just the system as a whole.

This will effectively fix the issue I raised in #377.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,6 +23,7 @@ Mahmoud Hamdy (TutTrue) <mahmoud.hamdy5113@gmail.com>
 Somye Mahajan <mahajan.somye@gmail.com>
 Shashank Singh <shashanksgh3@gmail.com>
 Abdelrhman Sersawy <abdelrhmansersawy@gmail.com>
+Benjamin Li <bli@linsang.com>
 Amr Shams (NightBird) <amr.shams2015.a@gmail.com>
 Mazen Kamal <mazenkamal212@gmail.com>
 Trevor Jacob Mathews <trevorjacobmathews@gmail.com>
--- a/contrib/json/postgresql-13.json
+++ b/contrib/json/postgresql-13.json
@@ -407,11 +407,20 @@
     {
       "tag": "pg_statio_all_tables",
       "collector": "statio_all_tables",
+      "database": "all",
       "queries": [
         {
-          "query": "SELECT COUNT(heap_blks_read) AS heap_blks_read, COUNT(heap_blks_hit) AS heap_blks_hit, COUNT(idx_blks_read) as idx_blks_read, COUNT(idx_blks_hit) AS idx_blks_hit, COUNT(toast_blks_read) AS toast_blks_read, COUNT(toast_blks_hit) AS toast_blks_hit, COUNT(tidx_blks_read) AS tidx_blks_read, COUNT(tidx_blks_hit) AS tidx_blks_hit FROM pg_statio_all_tables;",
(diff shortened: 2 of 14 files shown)


bassamadnan:
LGTM, 

a few considerations while we are on this topic to improve this query

1. maybe we can use  `COALESCE(idx_blks_read, 0)` I am not sure how this would behave for tables without indexes.
2. Why not filter `WHERE schemaname NOT IN ('pg_catalog', 'information_schema')`?

benjunmun:
> LGTM,
> 
> a few considerations while we are on this topic to improve this query
> 
>     1. maybe we can use  `COALESCE(idx_blks_read, 0)` I am not sure how this would behave for tables without indexes.

Some of those columns appear to be intentionally null when the value is not applicable. But as far as I can tell the pgexporter infrastructure replaces those with zeroes already.

> 
>     2. Why not filter `WHERE schemaname NOT IN ('pg_catalog', 'information_schema')`?

I could change this metric entirely to be based off of the `pg_statio_user_tables` view instead of `pg_statio_all_tables`. That pretty much does the same thing but communicates the intent to users. Should I make this change? I don't have an enormous amount of postgres admin experience, but I assume that the system tables are much less likely to be a source of IO concerns and therefore the user tables are more interesting metrics to include by default and avoid bloating the output.

bassamadnan:
> I could change this metric entirely to be based off of the `pg_statio_user_tables` view instead of `pg_statio_all_tables`. That pretty much does the same thing but communicates the intent to users. Should I make this change? I don't have an enormous amount of postgres admin experience, but I assume that the system tables are much less likely to be a source of IO concerns and therefore the user tables are more interesting metrics to include by default and avoid bloating the output.

Well, you're not wrong. Lets keep it as it is then.

bassamadnan:
@jesperpedersen LGTM, @benjunmun could you please rebase?

benjunmun:
> @jesperpedersen LGTM, @benjunmun could you please rebase?

Done!

jesperpedersen:
@benjunmun Are you planning updates to the Grafana dashboards, or do you see that as a separate issue ?

jesperpedersen:
@benjunmun Please, rebase
@bassamadnan PTAL

benjunmun:
I've rebased.

I hadn't thought about the Grafana dashboards. I can evaluate the changes there, but it'll take me longer to get the time to set up an environment for that and take a look. I can do that as part of this issue or a separate fix if you'd like to merge this now. Do you have a preference?

jesperpedersen:
@benjunmun I would do it as part of this pull request

jesperpedersen:
@benjunmun Rebase, and check the new infrastructure

jesperpedersen:
@benjunmun Can you rebase ?

bassamadnan:
@benjunmun any updates?

jesperpedersen:
@benjunmun Can you resolve this ?

benjunmun:
I've rebased.

Checked over the grafana dashboards, it's not obvious to me that anything needs to change? I've only fixed the pg_statio_all_tables and pg_statio_all_sequences metrics, and I don't see either of those referenced in the dashboards.

bassamadnan:
> I've rebased.
> 
> Checked over the grafana dashboards, it's not obvious to me that anything needs to change? I've only fixed the pg_statio_all_tables and pg_statio_all_sequences metrics, and I don't see either of those referenced in the dashboards.

That will be okay, thanks for your contribution ! :) @benjunmun

PR #375, opened 2026-01-25
[#374] Fix SSL leaks

bassamadnan:
Okay based on what I have read a few things are causing this issue

1. OpenSSL objects apparently can not be shared between processes, if I understand it correctly - we create connections and pass it down on forked processes to serve metrics. At times when the child process uses `pgexporter_switch_db` the SSL pointer is updated to the new connection, and exits after closing its own connections, the child process ends up overwriting the pointer which ends up orphaning the main process's SSL object. I read up on this and apparently keeping SSL objects in shared memory may not be a good idea. [[1](https://stackoverflow.com/questions/41323286/openssl-session-pool-in-shared-memory-shm-possible)] [[2](https://mta.openssl.org/pipermail/openssl-users/2022-September/015443.html)]. Though I may not understand the issue completely, a google overview on the search "can ssl connection be shared between processes" gives this reasoning: "..the active SSL connection itself cannot be easily shared or "transferred...as they contain internal memory buffers and pointers to other OpenSSL structures...". I may have misunderstood/may be wrong but introducing a `pgexporter_close_connections` in the `main` loop just before loading extensions `pgexporter_load_extension_yamls` gets rid of a few memory leaks after 2-3 cURLs, indicating that some of the connections were not properly being freed earlier. Further, as some of the comments earlier in other issues mentioned a `malloc` leak inside the OpenSSL paths, I suspect this is some internal pointer private to the process creating the SSL connection and never free-ing it. 
2. Further, `pgexporter_switch_db` disconnects a socket descriptor, but does not clean up/disconnect the ssl connection itself, adding that logic helped (i.e closing the ssl connection when switching to different databases) fixed memory leaks from there. We open connections anyway in `[metrics_page](https://github.com/pgexporter/pgexporter/blob/e0b5753199fcf29b683097a3adf4db6ca46c2a41/src/libpgexporter/prometheus.c#L673)` So we close those while switching to different databases.
3. Much earlier on [this](https://github.com/pgexporter/pgexporter/pull/210#issuecomment-2825023929) thread, we discussed about the same issue - i.e managing SSL connections, since I figured that SSL was causing some leaks when we cache it, due to it almost acting as a process-local object, i just free-ed it - forcing future process to create their own (also modified the open/close connection APIs to reflect this) and this solved the remaining leaks, this also removed " 2025-04-17 09:42:51 ERROR message.c:936 SSL_ERROR_SSL: Success (-1)" messages I would see on my logs back then in the thread I shared. 
[...]

PR #381, opened 2026-01-31
[#375 follow up] Fix integer overflow and add required permissions documentation

cracksalad:
This is a follow up of #375 

Got the following warnings about queries, that fail:
- pg_table_bloat: see below
- pg_file_settings: permission issue (`pg_file_settings` and `pg_show_all_file_settings`)
- pg_shadow: permission issue (`pg_authid`)
- pg_usr_evt_trigger: permission issue (`pg_authid`)

**About pg_table_bloat**
Got `ERROR:  integer out of range` error from that query. Via binary search, I quickly found the cause :-D and added a type cast to `bigint`.

___

@jesperpedersen Regarding the permission documentation: Would you prefer to replace "must" with something like "should" since it will trigger a warning but other than that, it does what it is supposed to?


jesperpedersen:
We should have the documentation in the manual (`doc/manual/en/`) as well

jesperpedersen:
@bassamadnan Can you take this one, and PTAL me once it is ready

bassamadnan:
> @bassamadnan Can you take this one, and PTAL me once it is ready

LGTM, no immediate issues i see on this

jesperpedersen:
Rebased, and merged.

Issue #389, opened 2026-02-03
remove hard coded content type when building a request


Amr-Shams:
**Is your feature request related to a problem? Please describe.**

`Content-Type` is currently hard-coded in `http_build_request()` based on the HTTP method (POST/PUT). This treats `Content-Type` as fixed behavior instead of a normal HTTP header and makes it impossible to reliably override or omit it.

**Describe the solution you'd like**

Treat `Content-Type` as a regular, optional header. Do not set it implicitly in the request builder, or only set a default if it is not already provided by the caller. or add it as a separate filed in the `request` (xml content type)

**Additional information**

`Content-Type` should be fully caller-controlled

PR #390, opened 2026-02-03
[#389] fet: removed hard coded content type when building http request

Amr-Shams:


jesperpedersen:
@Amr-Shams Authors, and you need to add the content type to existing responses, right ?

jesperpedersen:
@bassamadnan PTAL

Amr-Shams:
> @Amr-Shams Authors, and you need to add the content type to existing responses, right ?



i have looked to the entire codebase, but the only use cases where for get requests, as there is no payloads, i don't really care about the content type, maybe when used with post/put/delete or requests with body they(caller) have to set it manually for the content type

bassamadnan:
> > @Amr-Shams Authors, and you need to add the content type to existing responses, right ?
> 
> i have looked to the entire codebase, but the only use cases where for get requests, as there is no payloads, i don't really care about the content type, maybe when used with post/put/delete or requests with body they(caller) have to set it manually for the content type

Thats right. What about the test cases? they would naturally be using the default content-type before so maybe the might need a change

Amr-Shams:
> > > @Amr-Shams Authors, and you need to add the content type to existing responses, right ?
> > 
> > 
> > i have looked to the entire codebase, but the only use cases where for get requests, as there is no payloads, i don't really care about the content type, maybe when used with post/put/delete or requests with body they(caller) have to set it manually for the content type
> 
> Thats right. What about the test cases? they would naturally be using the default content-type before so maybe the might need a change

nop, still tests the get request, nothing in the codebase uses the content type, as far as now

bassamadnan:
okay, just a small typo in your commit message I believe it would've been "feat*" apart from that LGTM

jesperpedersen:
@Amr-Shams you don't need to add `feat:` - just the issue number

jesperpedersen:
Rebased, and merged.

PR #391, opened 2026-02-03
[#373] Added query execution statistics

### LINKED ISSUE #373: Query execution statistics
Introduce counters like `pgexporter_query_executions_total`, `pgexporter_query_errors_total`, `pgexporter_query_timeouts_total` to increase the visibility of the queries & their failures.

sommye-ctr:
Fixes #373 

### Changes
- Added `pgexporter_query_executions_total` `pgexporter_query_errors_total` `pgexporter_query_timeouts_total`

### DIFF
--- a/src/libpgexporter/queries.c
+++ b/src/libpgexporter/queries.c
@@ -43,7 +43,10 @@
 /* system */
 #include <stdlib.h>
 
+#define SQLSTATE_QUERY_CANCELED "57014"
+
 static int query_execute(int server, char* qs, char* tag, int columns, char* names[], struct query** query);
+static bool is_query_timeout_error(struct message* error_msg);
 static void* data_append(void* orig, size_t orig_size, void* n, size_t n_size);
 static int create_D_tuple(int server, int number_of_columns, struct message* msg, struct tuple** tuple);
 static int get_number_of_columns(struct message* msg);
@@ -660,6 +663,49 @@ pgexporter_get_column_by_name(char* name, struct query* query, struct tuple* tup
    return NULL;
 }
 
+static bool
+is_query_timeout_error(struct message* error_msg)
+{
+   bool is_timeout = false;
+   if (error_msg != NULL && error_msg->length > 5)
+   {
+      char* payload = (char*)error_msg->data;
+      size_t offset = 5; /* kind (1) + length (4) */
+
+      while (offset < error_msg->length)
+      {
+         char field_type = payload[offset];
+         if (field_type == '\0')
+         {
+            break;
+         }
+
+         char* value = pgexporter_read_string(payload + offset + 1);
+
+         if (field_type == 'C')
+         {
+            if (!strcmp(value, SQLSTATE_QUERY_CANCELED))
+            {
+               is_timeout = true;
+               break;
+            }
+         }
+         else if (field_type == 'M')
+         {
+            if (strstr(value, "statement timeout") != NULL || strstr(value, "canceling statement due to user request") != NULL)
+            {
+               is_timeout = true;
+               break;
+            }
+         }
+
+         offset += 1 + strlen(value) + 1;
+      }
+   }
+
+   return is_timeout;
+}
+
 static int
 query_execute(int server, char* qs, char* tag, int columns, char* names[], struct query** query)
 {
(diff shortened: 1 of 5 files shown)


jesperpedersen:
@bassamadnan PTAL

src/libpgexporter/queries.c
```diff
@@ -660,6 +661,58 @@ pgexporter_get_column_by_name(char* name, struct query* query, struct tuple* tup
    return NULL;
 }
 
+static bool
+is_query_timeout_error(void* data, size_t data_size)
+{
+   struct message* error_msg = NULL;
+   bool is_timeout = false;
+
+   if (pgexporter_extract_message_from_data('E', data, data_size, &error_msg) != 0 || error_msg == NULL)
```
bassamadnan: This is a fairly redundant check, since this path is being reached by already checking the first condition

src/libpgexporter/queries.c
```diff
@@ -660,6 +661,58 @@ pgexporter_get_column_by_name(char* name, struct query* query, struct tuple* tup
...
+         if (field_type == '\0')
+         {
+            break;
+         }
+
+         char* value = pgexporter_read_string(payload + offset + 1);
+
+         if (field_type == 'C')
+         {
+            if (!strcmp(value, "57014"))
```
bassamadnan: Using a macro or atleast a variable would be good here, I had to look up what 57014 is.

src/libpgexporter/queries.c
```diff
@@ -660,6 +661,58 @@ pgexporter_get_column_by_name(char* name, struct query* query, struct tuple* tup
...
+            {
+               is_timeout = true;
+               break;
+            }
+         }
+         else if (field_type == 'M')
+         {
+            if (strstr(value, "statement timeout") != NULL || strstr(value, "canceling statement due to user request") != NULL)
```
bassamadnan, line 695: just curious- why is the second one also being considered as a timeout? (canceling statement due to user request)

bassamadnan: also wanted to ask, how do you test this while development? (i.e if it works, bounds etc)

sommye-ctr: With the information I could gather

> `57014` indicates that a SQL statement was cancelled or terminated, usually due to a query timeout, resource limit violation, or user interruption

so the first check for the code should suffice most of the time. The second message check is rather a defensive one and the _canceling statement due to user request_ is covered within the 57104 SQLSTATE, so to capture that it is included in the check.

sommye-ctr: For checking, I added a test metric which sleeps for 1s and then added a query timeout of 50ms. The counters were being inflated correctly in it.

sommye-ctr:
@jesperpedersen I faced this issue when working on this PR and therefore writing it here - the internal metrics are not loaded when -Y flag is passed so is it the intended way? I thought there was something wrong going on due to this.

jesperpedersen:
@sommye-ctr When you are using JSON or YAML as input then `internal.h` is disabled - if that is what you mean

bassamadnan:
@sommye-ctr the CI checks are failing

sommye-ctr:
I have separated the query stats info into a new function which is called AFTER all the other queries are run so that latest stats are printed. Also fixed memory leak issue.

bassamadnan:
> I have separated the query stats info into a new function which is called AFTER all the other queries are run so that latest stats are printed. Also fixed memory leak issue.

Out of curiosity, do you have any clue why there were leaks? The `if data != NULL` seems abit new to me so I am not sure why that condition became necessary now

sommye-ctr:
In the `is_timeout_error` function, I removed the 
   `if (pgexporter_extract_message_from_data('E', data, data_size, &error_msg) != 0 || error_msg == NULL)` check without thinking but this was populating the `error_msg` variable, so the variable was never getting populated.
   
 Now, I am passing the error msg directly as an argument into the function since this function should be called when there is an error.
 
 For the `if data != NULL` condition, this is inside the prometheus metrics which I added in the last push - basically creating a new function for showing these query data at the end of the output (previously was shown along with logging metrics but since the logging metrics are printed first, the errors or timeouts which happen in other metric functions were not reflected in the query stats)

bassamadnan:
Okay, LGTM. Look out for more of these SQL state stuff in the future, having one variable is always abit awkward so if we find enough relevent reasons to add more it would be good.

sommye-ctr:
> Okay, LGTM. Look out for more of these SQL state stuff in the future, having one variable is always abit awkward so if we find enough relevent reasons to add more it would be good.

I'll keep it in mind, currently I don't see any additional states that need to be added for the functionality.

jesperpedersen:
@sommye-ctr Any updates to the Grafana dashboard, or separate issue ?

sommye-ctr:
@jesperpedersen I think creating a new issue for that would be better

jesperpedersen:
Rebased, and merged.

PR #394, opened 2026-02-06
[#194] implement time based configuration generalization

### LINKED ISSUE #194: Time based configuration generalization

Abdelrhmansersawy:
fix #194 

Introduce a new uniform time-based metric type `pgexporter_time_t` that stores time in **milliseconds**, which is the smallest time unit needed by pgexporter.

 Apply the new uniform time type to:
 * `blocking_timeout`
 * `metrics_cache_max_age`
 * `metrics_query_timeout`
 * `bridge_cache_max_age`
 * `log_rotation_age`
 * `authentication_timeout`

### DIFF
--- a/src/include/pgexporter.h
+++ b/src/include/pgexporter.h
@@ -162,6 +162,23 @@ extern "C" {
            __typeof__ (b) _b = (b);  \
            _a < _b ? _a : _b; })
 
+/**
+ * Duration type stored internally as milliseconds.
+ */
+typedef struct pgexporter_time
+{
+   int64_t ms; /**< Internal storage in milliseconds */
+} pgexporter_time_t;
+
+#define PGEXPORTER_TIME_MS(v)    ((pgexporter_time_t){.ms = (int64_t)(v)})
+#define PGEXPORTER_TIME_SEC(v)   ((pgexporter_time_t){.ms = (int64_t)(v) * 1000LL})
+#define PGEXPORTER_TIME_MIN(v)   ((pgexporter_time_t){.ms = (int64_t)(v) * 60000LL})
+#define PGEXPORTER_TIME_HOUR(v)  ((pgexporter_time_t){.ms = (int64_t)(v) * 3600000LL})
+#define PGEXPORTER_TIME_DAY(v)   ((pgexporter_time_t){.ms = (int64_t)(v) * 86400000LL})
+
+#define PGEXPORTER_TIME_DISABLED ((pgexporter_time_t){.ms = 0})
+#define PGEXPORTER_TIME_INFINITE ((pgexporter_time_t){.ms = -1})
+
 /*
  * Common piece of code to perform a sleeping.
  *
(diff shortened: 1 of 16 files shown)


Abdelrhmansersawy:
@jesperpedersen I have implemented draft proposed solution.
PTAL

sommye-ctr:
@Abdelrhmansersawy I dont think the metrics query timeout has a **default** value of 50ms, it is the minimum value that it supports

Abdelrhmansersawy:
> @Abdelrhmansersawy I dont think the metrics query timeout has a **default** value of 50ms, it is the minimum value that it supports

Thanks for this review, I fixed it

src/include/pgexporter.h
```diff
@@ -162,6 +162,38 @@ extern "C" {
...
+#define PGEXPORTER_TIME_MS(ms)   ((pgexporter_time_t)(ms))
+#define PGEXPORTER_TIME_SEC(s)   ((pgexporter_time_t)((s) * 1000LL))
+#define PGEXPORTER_TIME_MIN(m)   ((pgexporter_time_t)((m) * 60000LL))
+#define PGEXPORTER_TIME_HOUR(h)  ((pgexporter_time_t)((h) * 3600000LL))
+#define PGEXPORTER_TIME_DAY(d)   ((pgexporter_time_t)((d) * 86400000LL))
+
+#define PGEXPORTER_TIME_DISABLED ((pgexporter_time_t)0)
+#define PGEXPORTER_TIME_INFINITE ((pgexporter_time_t)(-1))
+
+static inline int64_t
```
jesperpedersen: We don't have `static inline` functions - it should be a general function in `utils.h`

jesperpedersen:
@bassamadnan PTAL

src/libpgexporter/utils.c
```diff
@@ -3350,3 +3350,21 @@ pgexporter_snprintf(char* buf, size_t n, const char* fmt, ...)
 
    return ret;
 }
+
+int64_t
+pgexporter_time_as_ms(pgexporter_time_t t)
```
bassamadnan: the arguments into these functions could be more descreptive of what the input is supposed to be, for eg `pgexporter_time_t time_in_ms` (something better can also be thought of) of that sort. Hope I am not confused needlessly, for eg the `typedef` does declare this: "Duration type stored as milliseconds (int64_t)." But I still think it would be easy to miss this while developing it.

bassamadnan:
While setting up the test cases too, maybe we can have a test case which checks the conf? Not sure how feasible it is, but i dont think current test cases cover this.

src/libpgexporter/utils.c
Abdelrhmansersawy: > the arguments into these functions could be more descreptive of what the input is supposed to be, for eg pgexporter_time_t time_in_ms (something better can also be thought of) of that sort. Hope I am not confused needlessly

I think renaming these to `pgexporter_time_t time_in_ms` would be confusing, since `pgexporter_time_t` is designed to abstract away time units and serve as a generalized time type across the entire codebase.

> for eg the typedef does declare this: "Duration type stored as milliseconds (int64_t)." But I still think it would be easy to miss this while developing it.

I was unsure whether to make it a `typedef` or a `struct`. From your response, I think using a `struct` would enforce its and help prevent developers from miss.

I would like to hear your opinion

bassamadnan: I honestly think `typedef struct pgexporter_time{...}pgexporter_time_t` wont be a bad idea, since it is followed elsewhere, any thoughts? @jesperpedersen

jesperpedersen: @Abdelrhmansersawy @bassamadnan I think the `struct` route is good - maybe have a helper function to convert into a specific representation. Think `FORMAT_TIME_MS`, `FORMAT_TIME_S`, `FORMAT_TIME_TIMESTAMP`, ...

Abdelrhmansersawy: I have a design in mind and wanted to check with you if it aligns with what you have in mind.

`int64_t pgexporter_time_format(pgexporter_time_t t, enum pgexporter_time_format_t fmt);`

```cpp
enum pgexporter_time_format_t {
   FORMAT_TIME_MS,   /**< Milliseconds */
   FORMAT_TIME_S,    /**< Seconds */
   FORMAT_TIME_MIN,  /**< Minutes */
   FORMAT_TIME_HOUR, /**< Hours */
   FORMAT_TIME_DAY,  /**< Days */
};
```

This could then be used as:
`pgexporter_time_format(config->authentication_timeout, FORMAT_TIME_S)`

Abdelrhmansersawy:
> While setting up the test cases too, maybe we can have a test case which checks the conf? Not sure how feasible it is, but i dont think current test cases cover this.

Just to confirm my understanding...
would it be okay to create a new `test_configuration` and add tests for the time methods?

src/libpgexporter/utils.c
jesperpedersen: I think we should change it a bit such that it follows the standard way, and be used for attributes as well, so

```
int
pgexporter_time_format(pgexporter_time_t t, enum pgexporter_time_format_t fmt, char** output);
```

With `FORMAT_TIME_TIMESTAMP` we likely need a UTF formatted string...

jesperpedersen:
@Abdelrhmansersawy Of course

jesperpedersen:
@Abdelrhmansersawy Although the time functions will likely live in `utils.h` so `test_utils`

src/libpgexporter/utils.c
```diff
@@ -3350,3 +3351,86 @@ pgexporter_snprintf(char* buf, size_t n, const char* fmt, ...)
...
+   if (str == NULL)
+   {
+      return 1;
+   }
+   memset(str, 0, 64);
+
+   if (fmt == FORMAT_TIME_TIMESTAMP)
+   {
+      /* Format as ISO 8601 UTC timestamp */
+      time_t secs = (time_t)(t.ms / 1000);
```
Abdelrhmansersawy, line 3402: is it enough for displaying `seconds`, or should I attach also `milliseconds`?

Abdelrhmansersawy:
For test cases, I have tested all functionality of time covering all possible scenario.

src/libpgexporter/utils.c
jesperpedersen: Yeah, include ms

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@Abdelrhmansersawy I think it is a good change, so could you port it to pgmoneta and pgagroal as well ?

Abdelrhmansersawy:
> @Abdelrhmansersawy I think it is a good change, so could you port it to pgmoneta and pgagroal as well ?

Sure, I’ll port it to pgmoneta and pgagroal as well.

jesperpedersen:
@Abdelrhmansersawy Thanks !

PR #395, opened 2026-02-08
Fix missing liburing-devel in RPM Dockerfile and source name in Docker build script

cracksalad:
This PR contains two fixes and a documentation extension for contrib/docker/rpm:
- `liburing-devel` was missing from RPM Dockerfile
- the expected source .tar.gz name in pgexporter.spec changed from `0.8.0.tar.gz` to `pgexporter-0.8.0.tar.gz` in [bd991cc](https://github.com/pgexporter/pgexporter/commit/bd991cc73a1950ca6968998285af1b08e10333b8#diff-02f1b9ab3afc3ca6014337df3bc29e0199f51f5cc82d061dd1b3be5fd7f43283L7-R7)
- add note to contrib/docker/rpm/README.md to rebuild the docker image for each RPM rebuild. Could probably be handled more elegantly, but a note in the README should be a sufficient first step


jesperpedersen:
Merged.

Issue #406, opened 2026-02-15
Add Multixact ID Age (datminmxid) metric


Abdelrhmansersawy:
`pg_db_vacuum` currently tracks `age(datfrozenxid)` but does not expose `datminmxid` age. Since Multixact wraparound can also stop the database, this creates a monitoring blind spot.

PR #407, opened 2026-02-15
[#406]: Add Multixact ID Age (datminmxid) metric

Abdelrhmansersawy:
Fix: #406 

Extend the current query to include `age(datminmxid)` instead of creating a new one, since it uses the same `pg_database` source, is related to `age(datfrozenxid)`, and avoids extra query overhead. 

**Screen shot**
I create stress-test to ensure it works fine.
Screenshot from 2026-02-15 14-50-20


jesperpedersen:
Rebased, and merged.

PR #414, opened 2026-02-18
[#410] implement internal pgexporter_alert

### LINKED ISSUE #410: Add built-in Prometheus alerting rules
pgexporter currently already exposes 1079 metrics across 49 collectors - but ships zero alerting rules.

We can provide ready-to-use alert rules for critical severity, e.g. `PgExporterDown`, `PostgreSQLDown`, `PostgreSQLReplicationLagCritical`, `PostgreSQLWALReceiverDown`, `PostgreSQLReplicationSlotInactive`, `PostgreSQLTransactionIDWraparound`, `PostgreSQLMultiXactWraparound`

Abdelrhmansersawy:
resolved #410 

## Changes

- Built a new internal alert system that evaluates server connection status and executes custom SQL queries alongside standard Prometheus metric collection.
- Added YAML configuration support (`alerts_path`) for defining custom alerts, and shipped pre-configured alerts out of the box, including `postgresql_down`, `connections_high`, and `replication_lag`.
- Enabled two alert modes:
  - `type: connection` - checks if a server is reachable
  - `type: query` - evaluates SQL query results against defined thresholds
- Allowed targeting alerts to server filter, so an alert can run exclusively on a primary or replica node.
- Exposed evaluation results as `pgexporter_alert_*` Prometheus metrics (1 = firing, 0 = healthy)
- Added a comprehensive C unit test suite (`test_alert.c`) to verify alert evaluation logic.
- Fully documented the alert system in `doc/ALERT.md`.

Side Note:
- The `threshold` property is now defined as an `int`. When configuring `type: query` alerts, use whole numbers for values or percentages (e.g., use `80` instead of `0.8`).
- Server connection checks are now lazy — pgexporter_connection_isvalid is only called when an alert actually needs it, and the result is cached. Benchmarks with ApacheBench show the overall overhead dropped from ~1.1s to ~0.01s.

## Screenshots

Alert message is fired to to slack config with Grafana using webhook
553258331-3422fa4b-a46e-4b4a-90f6-186f0658e46d


Abdelrhmansersawy:
For alert handling, should we start with a Grafana-side approach (pgexporter emits all alert metrics and users choose what to alert on), or add pgexporter-side configurability from the start (e.g., an `alerts=` option similar to `collectors`)?

Does this function end up running `X` separate server loops (one for each SQL-based alert), meaning `X×N` queries over the network for N servers - similar to how `version_information` and `uptime_information` work? Could this cause any noticeable database overhead?

jesperpedersen:
@Abdelrhmansersawy See my comments on the issue - this is a lot bigger

src/libpgexporter/yaml_configuration.c
```diff
@@ -1843,3 +1843,298 @@ semantics_extension_yaml(struct configuration* config, yaml_config_t* yaml_confi
...
+         case YAML_MAPPING_END_EVENT:
+            if (in_alert)
+            {
+               in_alert = 0;
+
+               if (strlen(current_alert.name) > 0)
+               {
+                  if (merge)
+                  {
+                     /* Merge: find existing alert by name and overwrite threshold */
```
Abdelrhmansersawy: Here I only track the operator and threshold in case of an override. I think users shouldn’t need to redefine the other default fields - they should only specify:

```
name: 
operator:
threshold:
```

WDYT?

jesperpedersen: @Abdelrhmansersawy You need to create an `alert_configuration.c` and load it as part of the `pgexporter` binary

src/libpgexporter/prometheus.c
```diff
@@ -736,6 +745,9 @@ metrics_page(SSL* client_ssl, int client_fd)
             goto error;
          }
 
+         /* Alert evaluation - must run before custom/extension metrics */
```
Abdelrhmansersawy: @jesperpedersen 
I faced an issue where the `postgresql_down` alert was not firing as intended. While debugging, I found that `alert_information` must run only after `pgexporter_open_connections()` establishes a valid `fd` for each server.
However, `custom_metrics()` and `extension_metrics()` call `pgexporter_switch_db()`, which temporarily sets `fd = -1` while reconnecting to a database. If this reconnection fails, `fd` remains `-1`, leading to a false `postgresql_down` alert even though the server is still running.

Abdelrhmansersawy:
**My latest changes**
1. I have added   -a, --alerts  Set the path to the alert file
2. I add server filter ('Both', 'primary', 'replica') for alert source
3. Add Test cases for basics stuff merge (override alert) logic
4. Also I tested `postgres_down` with slack, and it works as intended
Screenshot from 2026-02-22 23-06-14

src/libpgexporter/alert_configuration.c
```diff
@@ -0,0 +1,358 @@
...
+{
+   yaml_parser_t parser;
+   yaml_event_t event;
+   int done = 0;
+   int in_alert_list = 0;
+   int in_alert_item = 0;
+   char* current_key = NULL;
+   struct alert_definition current_alert;
+   int alert_idx;
+   bool has_threshold;
```
Abdelrhmansersawy: For override keys that users can change, I think these could include `threshold`, `operator`, and `filter`.
Should we also support overriding other fields such as `name`, `description`, `query`, or `type`?

jesperpedersen:
@Abdelrhmansersawy If you are going down the `INTERNAL_ALERTS_YAML` route then we will need an `alerts = on` option

src/libpgexporter/prometheus.c
jesperpedersen: @Abdelrhmansersawy You can only run alerts if there is a valid `fd`. See

```
bool
pgexporter_connection_isvalid(SSL* ssl, int socket)
```

which may can help you

src/libpgexporter/alert_configuration.c
jesperpedersen: @Abdelrhmansersawy Try and document the format. A user can write whatever alerts they want, so you have to have a format that can translate into a Prometheus metric. A user can also provide their own scraping YAML file, so you will need to account for that.

Sidenote, `servers: both` - there could be 3+ servers so it is `all`

Abdelrhmansersawy: @jesperpedersen 
> Sidenote, servers: both - there could be 3+ servers so it is all

Should be a single choice:

```
servers: all
servers: primary
servers: replica
```

Or a list of servers:

```
servers:
    - primary
    - replica
    - etc
```

jesperpedersen: a list - where `all` is an alias

Abdelrhmansersawy:
@jesperpedersen 
I have done implementing all options that you have required from me.

What do you see next step on this issue?
I think we need to ensure that's very light, any idea how to improve this?

jesperpedersen:
@Abdelrhmansersawy Please, rebase - and port to the new testing framework

Abdelrhmansersawy:
> @Abdelrhmansersawy Please, rebase - and port to the new testing framework

Done, for now I use private shmem approach.

Abdelrhmansersawy:
I have updated test cases to use MCTF shared config isolation after https://github.com/pgexporter/pgexporter/pull/447.

jesperpedersen:
@Abdelrhmansersawy Rebase

Abdelrhmansersawy:
> @Abdelrhmansersawy Rebase

Done, Ready for review.

doc/ALERT.md
```diff
@@ -0,0 +1,59 @@
...
+The `alerts_path` must be a single yaml file.
+
+Each yaml is a sequence of alerts. Every alert can include seven properties.
+
+## alerts yaml
+
+| Property | Default | Required | Description |
+|----------|---------|----------|-------------|
+| name | | Yes | The unique identifier of the alert |
+| description | | Yes\* | Human-readable HELP description. |
```
jesperpedersen: What is this `\*` ?

doc/ALERT.md
```diff
@@ -0,0 +1,59 @@
+# Customized Alerts configuration
+
```
jesperpedersen, line 2: We need an introduction - what are alerts ? how do you see them ? why would you use them ? ...

src/include/configuration.h
```diff
@@ -53,6 +53,7 @@ extern "C" {
 #define CONFIGURATION_ARGUMENT_BRIDGE_JSON_CACHE_MAX_SIZE "bridge_json_cache_max_size"
 #define CONFIGURATION_ARGUMENT_MANAGEMENT                 "management"
 #define CONFIGURATION_ARGUMENT_CACHE                      "cache"
+#define CONFIGURATION_ARGUMENT_ALERTS                     "alerts"
```
jesperpedersen, line 54: Sorting

doc/ALERT.md
```diff
@@ -0,0 +1,86 @@
...
+|----------|---------|----------|-------------|
+| name | | Yes | The unique identifier of the alert |
+| description | | Yes | Human-readable HELP description. |
+| type | | Yes | `connection` or `query`. |
+| query | | No | The SQL query to execute (only for `type: query`). Should return a single integer value. |
+| operator | `>` | No | The comparison operator (only for `type: query`). Valid options: `>`, `<`, `>=`, `<=`, `==`, `!=` |
+| threshold | | Yes | The integer threshold value to compare against (only for `type: query`). |
+| servers | `all` | No | Target servers for the alert. Can be `all`, a specific server name (e.g., `primary`), or a list of server names (`[primary, replica]`). |
+
+**Note:** The `threshold` property is defined as an integer — for example, `80` for 80% of `max_connections`, or `1500000000` for a transaction ID age limit. Queries that return a percentage should scale the result to a whole number, for example `SELECT (count(*) * 100 / max_conn) FROM ...` so that a threshold of `80` means 80%.
```
Abdelrhmansersawy, line 42: @jesperpedersen 
as follow-up, I see it's best to add a side-note about you should add `%` instead of `0.x` for percentage metric as threshold is stored as integer - avoid float precision problem and for more readability.

Abdelrhmansersawy:
@jesperpedersen PTAL - I have addressed all comments

jesperpedersen:
@Abdelrhmansersawy We need the alerts chapter in the manual as well

Abdelrhmansersawy:
@jesperpedersen 
I see that we should add `12-alerts.md` but this is required shifted `12-15` chapters.
is it ok?

jesperpedersen:
@Abdelrhmansersawy Yes

jesperpedersen:
Merged.

jesperpedersen:
@Abdelrhmansersawy This can be cross-ported - of course with the focus of the other projects main idea

Issue #416, opened 2026-02-18
Prometheus label parsing breaks on quoted values containing braces


sommye-ctr:
**Describe the bug**

Metrics with label values like "{plpgsql}" are parsed incorrectly.
An example of this is:

```
pgexporter_pg_installed_extensions{server="primary", extensions="{plpgsql}", database="postgres"} 1
pgexporter_pg_installed_extensions{server="replica", extensions="{plpgsql}", database="postgres"} 1
```
gets converted to 
`server=primary, extensions=primary, plpgsql=, database=, "=database=, database=postgres`

**To Reproduce**

Test with any metrics which contain such braces (example given above)

jesperpedersen:
@sommye-ctr That is def a bug - it should be `extensions="plpgsql"` instead

jesperpedersen:
@sommye-ctr Try and install more extensions and see how it goes

sommye-ctr:
I also found that spaces cause some distortion too. Example:
```
pgexporter_pg_stat_io_reads{server="primary", backend_type="background worker", database="postgres"} 0
pgexporter_pg_stat_io_reads{server="primary", backend_type="slotsync worker", database="postgres"} 0
pgexporter_pg_stat_io_reads{server="primary", backend_type="client backend", database="postgres"} 827
pgexporter_pg_stat_io_reads{server="primary", backend_type="standalone backend", database="postgres"} 546
```
gives something like 
```
server=primary, backend_type=background, worker"=database=, database=postgres
server=primary, backend_type=slotsync, worker"=database=, database=postgres
server=primary, backend_type=client, backend"=database=, database=postgres
```

jesperpedersen:
@sommye-ctr See if you can get to the root cause of these issues

sommye-ctr:
I think its `prometheus_client.c`, in it currently `{`, `}`, `,` and space are used as delimiters without any awareness of whether these are inside quotes or not which gives this error. A solution would be to make a quote aware parsing logic

PR #417, opened 2026-02-19
[#416] Fixed Prometheus label parsing

sommye-ctr:
### Changes
- Replaced delimiter based tokenisation with quote aware parsing
- Enabled support for special characters in values (like `{}`)
- Added escape sequence and whitespaces handling

### DIFF
--- a/contrib/json/postgresql-13.json
+++ b/contrib/json/postgresql-13.json
@@ -204,7 +204,7 @@
       "collector": "installed_extensions",
       "queries": [
         {
-          "query": "SELECT array_agg(extname) AS extensions, count(*) FROM pg_extension;",
+          "query": "SELECT extname AS extensions, 1 AS installed FROM pg_extension;",
           "version": 10,
           "columns": [
             {
--- a/contrib/json/postgresql-14.json
+++ b/contrib/json/postgresql-14.json
@@ -204,7 +204,7 @@
       "collector": "installed_extensions",
       "queries": [
         {
-          "query": "SELECT array_agg(extname) AS extensions, count(*) FROM pg_extension;",
+          "query": "SELECT extname AS extensions, 1 AS installed FROM pg_extension;",
           "version": 10,
           "columns": [
             {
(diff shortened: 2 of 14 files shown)


jesperpedersen:
@sommye-ctr My point was we should never have the `{`/`}` in the first place inside an attribute... So, the bug must be somewhere else

sommye-ctr:
@jesperpedersen Oh.. I thought pgexporter's prometheus service design choice was made that way. Anyhow, the spaces between quotes bug is fixed by this. 

I'll investigate the source for the braces then.

jesperpedersen:
@sommye-ctr The important part is that Prometheus can parse it

sommye-ctr:
Afterall its a string so I think it should be able to do that. 
The postgres format for array is being directly parsed - so for this then we would have to expand our logic to emit one sample per array element.

sommye-ctr:
Or I think better fix is to then remove wherever `array_agg` is used in queries

jesperpedersen:
@sommye-ctr Just remember that we can only have a single key. My initial thought is that it is best to remove the `{`/`}` for now. Please, work towards a solution, and present how the actual Prometheus line would look with multiple extensions installed

sommye-ctr:
The aggregation behaviour is coming from 
`SELECT array_agg(extname) AS extensions, count(*) FROM pg_extension;`
This can be changed to 
`SELECT extname AS extensions, 1 AS installed FROM pg_extension;`
in the `INTERNAL_YAML` to fix this and the resultant output would be 

```
pgexporter_pg_installed_extensions{server="primary", extensions="plpgsql", database="postgres"} 1
pgexporter_pg_installed_extensions{server="primary", extensions="hstore", database="postgres"} 1
pgexporter_pg_installed_extensions{server="replica", extensions="plpgsql", database="postgres"} 1
pgexporter_pg_installed_extensions{server="replica", extensions="hstore", database="postgres"} 1
```
Should I proceed to patch this then?

jesperpedersen:
@sommye-ctr Yes, see if `pgexporter_postgresql_extension_info` need changes as well

sommye-ctr:
> @sommye-ctr Yes, see if `pgexporter_postgresql_extension_info` need changes as well


It works fine without any changes to it

jesperpedersen:
Rebased, and merged.

PR #426, opened 2026-02-21
[#425] Fix missing SIGCHLD handler causes zombie process leak

### LINKED ISSUE #425: Missing SIGCHLD handler causes zombie process leak on fork() commands
The daemon uses fork() but never reaps its terminated child processes, resulting in zombie processes accumulating over time.

as SIGCHLD is missing from the event loop and there are no waitpid() calls, every ephemeral collector process that finishes execution is left behind in the process table as a zombie

fix:

port the sigchld_cb logic from pgagroal

trxvorr:
fixes #425

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -72,6 +72,7 @@
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Rebased, and merged.

PR #428, opened 2026-02-21
[#427] Update check.sh to run tests from the project directory

### LINKED ISSUE #427: test_database_extension_path test fails when check.sh is run from the build/ directory
**Describe the bug**

The `test_database_extension_path` test depends on the execution working directory. It fails with **"Extension path is empty"** when the test suite is run from inside the `build/` directory (`cd build && ./check.sh`), but succeeds when executed from the project root (`./build/check.sh`).

**To Reproduce**

Build the project:

```bash
cd pgexporter
mkdir build
cd build
cmake -DCMAKE_C_COMPILER=clang -DCMAKE_BUILD_TYPE=Debug ..
make
```

Run the test suite from the `build/` directory:

```bash
./check.sh
```

Abdelrhmansersawy:
Fix #427

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -289,6 +289,8 @@ execute_testcases() {
    fi
 
    echo "Start running tests"
+   # Run tests from the project root
+   cd "$PROJECT_DIRECTORY"
    $TEST_DIRECTORY/pgexporter-test
    if [[ $? -ne 0 ]]; then
       exit 1


jesperpedersen:
@Abdelrhmansersawy I don't know about this one, I run it as `./test/check.sh`
@shashidhar-bm Thoughts ?

shashidhar-bm:
@jesperpedersen , I think it’s needed when running from the build dir (e.g. `cd build && ./check.sh`). The other option would be path handling in `test_database_extension_path`, but I feel this approach is the better one

jesperpedersen:
@shashidhar-bm My point was maybe we need to detect where we are calling `check.sh` from. F.ex. I do `./test/check.sh` from top-level directory after building the project

shashidhar-bm:
@jesperpedersen Ah, that makes sense — it adds clarity and keeps the code cleaner.
(also with the current change it's a no-op when run from the top level)

Abdelrhmansersawy:
> My point was maybe we need to detect where we are calling `check.sh` from. F.ex. I do `./test/check.sh` from top-level directory after building the project

Yes, the correct place is the top-level, but contributors might run by mistakes it from somewhere else (e.g. `/src/...`). Maybe we can make this easier by adding `cd "$PROJECT_DIRECTORY"` in `check.sh` before running the test cases to avoid issues when it’s executed from the wrong location.

Abdelrhmansersawy:
It actually happened to me while trying to create a shell function to build and test the project with single command :D
```fish
function pg_build
    # if inside build/, go back
    if test (basename (pwd)) = "build"
        cd ..
    end

    # ensure build directory exists
    mkdir -p build
    cd build

    # configure + build
    cmake -DCMAKE_C_COMPILER=clang -DCMAKE_BUILD_TYPE=Debug ..
    make

    # run checks
    chmod +x check.sh
    ./check.sh
end
```

jesperpedersen:
Rebased, and merged.

jesperpedersen:
@shashidhar-bm Cc: ^

PR #443, opened 2026-02-25
[#441] Generate Version-Specific Grafana Dashboards

### LINKED ISSUE #441: Generate Version-Specific Grafana Dashboards from a Single Source
I was thinking about why we don’t keep a single Grafana `postgresql_dashboard.json` and generate version-specific dashboards from it - similar to what we already do with `yaml/postgresql-xx.yaml`, which are generated from `Internal.h` using a Python script.

I think we could follow the same idea for dashboards.

I looked through the different Grafana dashboard versions to see if this is possible, and it seems we mainly just need to know the minimum PostgreSQL version required for each panel. e.g:
- `WAL Generation` - pg14+
- `Memory Usage by Context` - pg14+
- `Wait Events by Type` - pg17+

This way we can keep a single source of truth and generate the dashboards automatically instead of maintaining multiple copies.

Abdelrhmansersawy:
fix #441

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,15 @@ jobs:
 
       - name: Check contrib file drift
         run: |
-          python3 scripts/generate_contrib.py \
+          python3 scripts/generate_yaml_json.py \
             --internal-h src/include/internal.h \
+            --check
+        working-directory: /home/runner/work/pgexporter/pgexporter/
+
+      - name: Check Grafana dashboard drift
+        run: |
+          python3 scripts/generate_dashboards.py \
+            --template contrib/grafana/postgresql_dashboard.json \
             --check
         working-directory: /home/runner/work/pgexporter/pgexporter/
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,15 +233,30 @@ endif()
 # Custom target to regenerate contrib files
 #
 if(Python3_FOUND)
-  set(GENERATOR_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/scripts/generate_contrib.py")
+  set(YAML_JSON_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/scripts/generate_yaml_json.py")
+  set(DASHBOARD_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/scripts/generate_dashboards.py")
   set(INTERNAL_H "${CMAKE_CURRENT_SOURCE_DIR}/src/include/internal.h")
-  
+  set(DASHBOARD_TEMPLATE "${CMAKE_CURRENT_SOURCE_DIR}/contrib/grafana/postgresql_dashboard.json")
+
   add_custom_target(
-    generate-contrib
-    COMMAND ${Python3_EXECUTABLE} ${GENERATOR_SCRIPT}
+    generate-yaml-json
+    COMMAND ${Python3_EXECUTABLE} ${YAML_JSON_SCRIPT}
             --internal-h ${INTERNAL_H}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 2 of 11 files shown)


jesperpedersen:
@Abdelrhmansersawy @bassamadnan will work with you on this one

Abdelrhmansersawy:
@bassamadnan 
Hi, PTAL.

Thanks !

bassamadnan:
Thanks for the PR! Will take a look soon, in middle of exams right now.

Abdelrhmansersawy:
> Thanks for the PR! Will take a look soon, in middle of exams right now.

Good luck with your exams !

bassamadnan:
> > Thanks for the PR! Will take a look soon, in middle of exams right now.
> 
> Good luck with your exams !

Thank you. I also wanted to ask, how do you test it for different postgres versions?

Abdelrhmansersawy:
> Thank you. I also wanted to ask, how do you test it for different postgres versions?

Currently, each metric inside `internal.h` has a `version` attribute which represents the minimum Postgres version required.

The same thing can be done by the developer on the Grafana dashboard. If they want some panels to be supported from `pg_x` version, they can add `"version": 14`.

e.g.

```
"title": "Memory Usage by Context",
"type": "timeseries",
"version": 14
```

Then my script detects this and only puts these panels on the correct versions.

bassamadnan:
@Abdelrhmansersawy I meant checking different versions, because often what would happen is the panel would sometimes just be empty. If you could provide some screenshots from the different postgres versions it would be good. My particular doubt is for the recalculating grid position function, I dont think our grid positions are uniform enough that the recalculation will still make it look pretty/not overlap, i might be wrong though

Abdelrhmansersawy:
> If you could provide some screenshots from the different postgres versions it would be good.

If you take a look on changed lines, you can see that my script didn't change any position of already existing panels.

**Here's example of panel version - PG18**
Screenshot from 2026-03-04 00-38-39

**Here's example of the same panel - PG13**
Screenshot from 2026-03-04 00-39-04

Abdelrhmansersawy:
Due to overlapping doubt, I will explain how I recalculate the panel's position after some removed panel:
it basically shifts gridPos.y values upward to fill gaps left by removed panels, no panels order changes or swap occur.

**Here's an illustration** 
image

So it's mathematically , enforce no overlapping happen.

Abdelrhmansersawy:
> I dont think our grid positions are uniform enough that the recalculation will still make it look pretty/

Of course, we can't 100% ensure an automatically polished layout without the developer manually choosing the position of the versioned panel in `postgresql_dashboard.json` to make it look good, using the same shifting logic.

You might think now this approach is wrong - but why? The developer will go through the same process when removing a panel from a dashboard version, so it’s the same thing.

My approach enforces the same order of panels layout, so it wouldn't be annoying to user when switching between dashboard version as we enforce this.

bassamadnan:
> If you take a look on changed lines, you can see that my script didn't change any position of already existing panels.

Yes, I  had noticed.
 
> **Here's example of panel version - PG18** 
> **Here's example of the same panel - PG13**

These look good and as expected.



> > I dont think our grid positions are uniform enough that the recalculation will still make it look pretty/
> 
> Of course, we can't 100% ensure an automatically polished layout without the developer manually choosing the position of the versioned panel in `postgresql_dashboard.json` to make it look good, using the same shifting logic.

That is right.

> You might think now this approach is wrong - but why? The developer will go through the same process when removing a panel from a dashboard version, so it’s the same thing.

That is not what I thought, I had just forgotten how Grafana layouts  were handeled via the JSON imports. 

LGTM, please ping me after a rebase. Took some time to get a panel counts and the update meta data API in python. @Abdelrhmansersawy

Abdelrhmansersawy:
@bassamadnan 
rebased Done.

Thanks for review !

bassamadnan:
> Thanks for review !
Thank you for your contribution !

PR #470, opened 2026-03-04
[#453] implement exclude collectors

### LINKED ISSUE #453: Disable whole collector filter
While working on optimizing the performance of alert system in [https://github.com/pgexporter/pgexporter/issues/410#issuecomment-3946997523](https://github.com/pgexporter/pgexporter/issues/410#issuecomment-3946997523), I need a way to disable some collectors from exposing metrics.

The pgexporter endpoint is currently very loaded, and I want to reduce the number of exported metrics so I can measure performance changes without pgexporter itself becoming the bottleneck.

I noticed that currently pgexporter does not support something like this. I looked at other monitoring systems and noticed they support something like:
`disable_collectors = x, y, z`

Abdelrhmansersawy:
resolved #453 

## The changes
1. Support `-X, --exclude-collectors` flag
2. Add `exclude_collectors` , `collectors` inside config file


jesperpedersen:
Rebased, and merged.

PR #474, opened 2026-03-09
[#473] Added auto-refresh button in web console

### LINKED ISSUE #473: Add option to auto-refresh web console after x duration
This should be done within the UI itself.
One example of the ui could be such:

Image

Manual refresh when clicked and option to set duration when clicked on right arrow.

sommye-ctr:
Resolves #473

Default option is set to 0 and it is enabled only when user enables it.

image
image


jesperpedersen:
Merged.

PR #483, opened 2026-03-21
Fix configuration segfault and assert() side-effects in test suite

trxvorr:


jesperpedersen:
Merged.

Issue #481, opened 2026-03-19
getenv crash


jesperpedersen:
2026-03-19 03:46:24 INFO  main.c:1323 pgexporter: shutdown
AddressSanitizer:DEADLYSIGNAL
=================================================================
==715554==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000b12 (pc 0x7fcbe0c2a88d bp 0x7fff9a85f600 sp 0x7fff9a85f5d0 T0)
==715554==The signal is caused by a READ memory access.
==715554==Hint: address points to the zero page.
    #0 0x7fcbe0c2a88d in getenv (/lib64/libc.so.6+0x1d88d) (BuildId: ff0267465bc3d76e21003b3bc5598fd5ee63e261)
    #1 0x5570b9aca6ba in writeFile InstrProfilingFile.c.o
    #2 0x5570b9aca476 in __llvm_profile_write_file (build/src/pgexporter+0x104476) (BuildId: ef28cf2604361940f8872ac461138f5a0d952528)
    #3 0x7fcbe0c299e0 in __run_exit_handlers (/lib64/libc.so.6+0x1c9e0) (BuildId: ff0267465bc3d76e21003b3bc5598fd5ee63e261)
    #4 0x7fcbe0c29abd in exit (/lib64/libc.so.6+0x1cabd) (BuildId: ff0267465bc3d76e21003b3bc5598fd5ee63e261)
    #5 0x7fcbe0c105bb in __libc_start_call_main (/lib64/libc.so.6+0x35bb) (BuildId: ff0267465bc3d76e21003b3bc5598fd5ee63e261)
    #6 0x7fcbe0c10667 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3667) (BuildId: ff0267465bc3d76e21003b3bc5598fd5ee63e261)
    #7 0x5570b99c6dd4 in _start (build/src/pgexporter+0xdd4) (BuildId: ef28cf2604361940f8872ac461138f5a0d952528)

==715554==Register values:
rax = 0x0000000000000013  rbx = 0x0000000000000b12  rcx = 0x00007fcbe0cf4e1b  rdx = 0x00005570b9b24b8f  
rdi = 0x00005570b9b24b8f  rsi = 0x000000000000002c  rbp = 0x00007fff9a85f600  rsp = 0x00007fff9a85f5d0  
[...]

PR #487, opened 2026-03-22
[#481] Fix environ pointer seg fault

bassamadnan:
pgexporter_free_proc_title() was freeing the heap copy of environ while the global environ pointer still referenced it. Any getenv() call after that point (such as from atexit handlers, i think by LLVM or ASAN injections)  would dereference freed memory, causing a segfault on shutdown.                                                                                                                           

Fix by not freeing the heap copy to remove the error- environ remains valid for the lifetime of the process and the OS reclaims the memory on exit anyway. Addresses #481

### DIFF
--- a/src/libpgexporter/utils.c
+++ b/src/libpgexporter/utils.c
@@ -3533,11 +3533,6 @@ pgexporter_free_proc_title(void)
    {
       return;
    }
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@bassamadnan Go ahead with merge

PR #492, opened 2026-03-25
[#491] Support storing PostgreSQL type OIDs for NULL values as default values.

### LINKED ISSUE #491: Null pointer passed to strcmp in add_column_to_store
**Describe the bug**

When running pgexporter, a runtime error is triggered at `prometheus.c:2103` inside the `add_column_to_store` function:

```
prometheus.c:2103:101: runtime error: null pointer passed as argument 2,
which is declared to never be null
/usr/include/string.h:172:33: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior prometheus.c:2103:101
```

Abdelrhmansersawy:
resolved #491

### DIFF
--- a/src/include/queries.h
+++ b/src/include/queries.h
@@ -55,6 +55,7 @@ struct query
    char tag[PROMETHEUS_LENGTH];                          /**< The tag */
    char names[MAX_NUMBER_OF_COLUMNS][PROMETHEUS_LENGTH]; /**< The column names */
    int number_of_columns;                                /**< The number of columns */
+   int type_oids[MAX_NUMBER_OF_COLUMNS];                 /**< The PostgreSQL type OIDs */
 
    struct tuple* tuples; /**< The tuples */
 } __attribute__((aligned(64)));
--- a/src/libpgexporter/prometheus.c
+++ b/src/libpgexporter/prometheus.c
@@ -193,6 +193,7 @@ static int parse_list(char* list_str, char** strs, int* n_strs);
 static char* get_value(char* tag, char* name, char* val);
 static int safe_prometheus_key_additional_length(char* key);
 static char* safe_prometheus_key(char* key);
+static char* safe_prometheus_attribute(char* attr, int type_oid);
 static void safe_prometheus_key_free(char* key);
 
 static bool is_metrics_cache_configured(void);
@@ -1942,9 +1943,27 @@ add_column_to_store(column_store_t* store, int store_idx, char* data, int sort_t
       {
          while (temp->next)
          {
-            if (temp->next->tuple != NULL && current != NULL && !strcmp(temp->next->tuple->data[0], current->data[0]))
+            if (temp->next->tuple != NULL && current != NULL)
             {
-               break;
+               char* next_d0 = temp->next->tuple->data[0];
+               char* cur_d0 = current->data[0];
+
+               if (next_d0 == NULL && cur_d0 == NULL)
+               {
+                  // both are NULL, group them together
+                  break;
+               }
+               else if (next_d0 == NULL && cur_d0 != NULL)
+               {
+                  // insert non-NULL before NULL, NULLs go last
+                  break;
+               }
+               else if (next_d0 != NULL && cur_d0 != NULL && !strcmp(next_d0, cur_d0))
+               {
+                  // both are non-NULL and equal, group them together
+                  break;
+               }
+               // cur is NULL but next is not, keep searching to put NULL last
             }
             temp = temp->next;
          }
@@ -2080,7 +2099,8 @@ append:
                   db_key_present = true;
                }
 
-               safe_key = safe_prometheus_key(pgexporter_get_column(j, current));
+               safe_key = safe_prometheus_attribute(pgexporter_get_column(j, current),
+                                                    temp->query->type_oids[j]);
                data = pgexporter_vappend(data, 5,
                                          ", ",
                                          temp->query_alt->node.columns[j].name,
@@ -2121,7 +2141,8 @@ append:
                db_key_present = true;
             }
 
-            safe_key = safe_prometheus_key(pgexporter_get_column(j, current));
+            safe_key = safe_prometheus_attribute(pgexporter_get_column(j, current),
+                                                 temp->query->type_oids[j]);
             data = pgexporter_vappend(data, 5,
                                       ", ",
                                       temp->query_alt->node.columns[j].name,
@@ -2162,7 +2183,8 @@ append:
                db_key_present = true;
             }
 
-            safe_key = safe_prometheus_key(pgexporter_get_column(j, current));
+            safe_key = safe_prometheus_attribute(pgexporter_get_column(j, current),
+                                                 temp->query->type_oids[j]);
             data = pgexporter_vappend(data, 5,
                                       ", ",
                                       temp->query_alt->node.columns[j].name,
@@ -2203,7 +2225,8 @@ append:
                db_key_present = true;
             }
 
-            safe_key = safe_prometheus_key(pgexporter_get_column(j, current));
+            safe_key = safe_prometheus_attribute(pgexporter_get_column(j, current),
+                                                 temp->query->type_oids[j]);
             data = pgexporter_vappend(data, 5,
                                       ", ",
                                       temp->query_alt->node.columns[j].name,
@@ -2620,6 +2643,16 @@ append:
 
          while (tuple)
          {
+            /* Skip tuples with NULL metric values */
+            char* metric_val = pgexporter_get_column(i, tuple);
+            if (metric_val == NULL)
+            {
+               pgexporter_log_debug("NULL metric value for %s_%s, skipping tuple",
+                                    store[idx].tag, store[idx].name);
+               tuple = tuple->next;
+               continue;
+            }
+
             data = NULL;
 
             data = pgexporter_vappend(data, 2,
@@ -2651,7 +2684,8 @@ append:
                   db_key_present = true;
                }
 
-               safe_key = safe_prometheus_key(pgexporter_get_column(j, tuple));
+               safe_key = safe_prometheus_attribute(pgexporter_get_column(j, tuple),
+                                                    temp->query->type_oids[j]);
                data = pgexporter_vappend(data, 5,
                                          ", ",
                                          temp->query_alt->node.columns[j].name,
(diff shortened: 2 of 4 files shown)


jesperpedersen:
@Abdelrhmansersawy Maybe, but what is the root cause ? We shouldn't have `NULL`s

Abdelrhmansersawy:
@jesperpedersen 
Maybe is the root cause:
https://github.com/pgexporter/pgexporter/blob/35d9f20363cac3e279dbff2012b2ccd60505de46/src/libpgexporter/queries.c#L925-L929

jesperpedersen:
@Abdelrhmansersawy Please, verify that is the case - you can add a TRACE logging statement showing the tuple information

Abdelrhmansersawy:
@jesperpedersen 
I traced the logs and found this:

```
DEBUG GC: tag=pg_shmem_allocations_numa sort=1 srv=0 d0=0x7b5d2caa0530
DEBUG ENTRY add_column_to_store: sort_type=1 current=0x7b9d2ca37320
DEBUG add_column_to_store: next->tuple=0x7b9d2ca37200 current=0x7b9d2ca37320 next->data[0]=0x7b5d2caa0530 current->data[0]=(nil)
```

The first column is `numa_node`, and `pg_shmem_allocations_numa` can return `NULL` for it. When that happens, PostgreSQL sends length `-1`, so `create_D_tuple` sets `data[0] = NULL`.

The issue is we’re sorting with `SORT_DATA0` (`strcmp` on `data[0]`), which crashes when `data[0]` is `NULL`.

Abdelrhmansersawy:
After fixing this, I ran pgexporter again and it worked fine.

jesperpedersen:
@Abdelrhmansersawy So, we need to be able to sort `NULL` values, and - at least - highlight them in the log. Maybe put the `NULL` last in sorting

Abdelrhmansersawy:
@jesperpedersen PTAL

Abdelrhmansersawy:
Now, Null values are put on the last of sort.
```
#HELP pgexporter_pg_shmem_allocations_numa_allocations Number of shared memory allocations on this NUMA node
#TYPE pgexporter_pg_shmem_allocations_numa_allocations gauge
pgexporter_pg_shmem_allocations_numa_allocations{server="primary", numa_node="0", database="postgres"} 75
pgexporter_pg_shmem_allocations_numa_allocations{server="primary", numa_node="", database="postgres"} 75
```

jesperpedersen:
@Abdelrhmansersawy Ok, but the attribute can't be empty... so we will need to filter it out, or make a _total. Lets start with former, although it isn't completely correct

jesperpedersen:
@sommye-ctr Can you keep an eye on this as well

Abdelrhmansersawy:
> @Abdelrhmansersawy Ok, but the attribute can't be empty... so we will need to filter it out, or make a _total. Lets start with former, although it isn't completely correct

What about set these attributes to `n/a`. e.g.
we already use this.

```
sersawy@omarchy ~ [0|1]> curl -s http://localhost:5002/metrics | grep '="n/a"'
pgexporter_pg_process_idle_seconds_bucket{le="+Inf", server="primary", application_name="n/a", database="postgres"} 0
pgexporter_pg_db_conn_ssl{server="primary", database="n/a"} 0
...
```

jesperpedersen:
@Abdelrhmansersawy Will `n/a` parse in this case ? I mean as a number... The thing is what does it mean ?
@sommye-ctr @bassamadnan  @resyfer We need a very strict parser for our test suite

Abdelrhmansersawy:
@Abdelrhmansersawy 

> Will n/a parse in this case ? I mean as a number... The thing is what does it mean ?
make sense

So, now do you think it’s enough to filter out null values (attribute value or metric value), and should I remove the sort handling for nulls since we don’t need it?

jesperpedersen:
@Abdelrhmansersawy Keep the sorting, but filter -- it is a much bigger issue

Abdelrhmansersawy:
Now, all empty attributes/values are filtered out

Before:
```
sersawy@omarchy ~> curl -s http://localhost:5002/metrics | grep '=""'
pgexporter_pg_replication_slots_active{server="primary", slot_name="repl", slot_type="physical", database="", two_phase_at="", invalidation_reason=""} 0
pgexporter_pg_replication_slots_temporary{server="primary", slot_name="repl", slot_type="physical", database="", two_phase_at="", invalidation_reason=""} 0
...
```

After:
```
sersawy@omarchy ~ [0|1]> curl -s http://localhost:5002/metrics | grep '=""'
sersawy@omarchy ~ [0|1]>
```

sommye-ctr:
> What about set these attributes to `n/a`. e.g.
> we already use this.

The "n/a" we are injecting right now is when an empty response is received (0 rows). In this issue, SQL NULL is being received.
For eg.
```
pgexporter_pg_replication_slots_active{server="primary", slot_name="n/a", slot_type="n/a", database="n/a"} 0
```
image

Ideally we should drop those queries from output which returns 0 rows. 



> Will `n/a` parse in this case ? I mean as a number... The thing is what does it mean ?

For the SQL NULL case - if the label is null, IMO we should keep it as 'none', 'n/a' or similar. Keeping it empty would lead to creation of new series if that value becomes non-null later on.

If the metric value is null, it is trickier. Maybe Nan? Maybe omit?

bassamadnan:
> If the metric value is null, it is trickier. Maybe Nan? Maybe omit?

Is omit same as dropping it? 

One way I potentailly see this being implemented is in the query result we can add a field for data being present, however cases where we get a number or text would have to be handled somehow at a query level, which I am not entirely sure how it would be implemented

sommye-ctr:
> Is omit same as dropping it?

Yes, I meant dropping it. 

> One way I potentailly see this being implemented is in the query result we can add a field for data being present, however cases where we get a number or text would have to be handled somehow at a query level, which I am not entirely sure how it would be implemented

I am not sure if I got it - like adding a 'status' label? 
`pg_table_size{table="users", status="absent"} 0`

jesperpedersen:
All keys must have the same attributes which has the same data type

jesperpedersen:
I think there is more to this...

Having

```
pgexporter_pg_replication_slots_active{server="primary", slot_name="repl", slot_type="physical", database="", two_phase_at="", invalidation_reason=""} 0
pgexporter_pg_replication_slots_temporary{server="primary", slot_name="repl", slot_type="physical", database="", two_phase_at="", invalidation_reason=""} 0
```

are important to have.

If we get a `NULL` we need to insert a default value based on the data type of the column. 

So,

```
pgexporter_pg_replication_slots_active{server="primary", slot_name="repl", slot_type="physical", database="n/a", two_phase_at="n/a", invalidation_reason="n/a"} 0
pgexporter_pg_replication_slots_temporary{server="primary", slot_name="repl", slot_type="physical", database="n/a", two_phase_at="n/a", invalidation_reason="n/a"} 0
```

Maybe we need something else for the default LSN value...

jesperpedersen:
For the above example - https://www.postgresql.org/docs/current/view-pg-replication-slots.html

Abdelrhmansersawy:
@jesperpedersen 
I’m thinking of doing it like this:

1. Extract the type OIDs from the T message and store them in the query struct - it's just simple.
```
int type_oid = pgexporter_read_int32(msg->data + offset + strlen(name) + 1 + 4 + 2);
```

2. `Keep data[i] = NULL` as is in `create_D_tuple`.

3. Then in `prometheus.c`, while building the output:
   - If column is `LABEL_TYPE` and `data[i] is NULL` - substitute based on type OID ("n/a", "0/0", etc.)
   - If column is `GAUGE/COUNTER` and data[i] is NULL - use "0" (or skip)

For label columns, we can map default values based on the type like this:
```c
 switch (type_oids[i]) {
        case 25:   // text
        case 19:   // name
        case 1043: // varchar
            result->data[i] = strdup("n/a");
            break;
        case 3220: // pg_lsn
            result->data[i] = strdup("0/0");
            break;
        default:
            result->data[i] = strdup("0");
            break;
```

jesperpedersen:
@Abdelrhmansersawy Yes, that sounds correct - `strdup` -> `pgexporter_append`. 

Maybe an idea is `static char* safe_prometheus_attrbute(char* attr);` and stick the `NULL` check in there...

sommye-ctr:
> * If column is `GAUGE/COUNTER` and data[i] is NULL - use "0" (or skip)

Using 0 wouldn't be wise - it will basically reset the counter and also skew the stats for the user.

Abdelrhmansersawy:
@sommye-ctr 
I think we can filter out it (skip this metrics scrape) and log this for user.

We can define skip_metric_scraps_total as counter that count it.

WDYT?

sommye-ctr:
> @sommye-ctr I think we can filter out it (skip this metrics scrape) and log this for user.
> 
> We can define skip_metric_scraps_total as counter that count it.
> 
> WDYT?

Seems good to me

Abdelrhmansersawy:
@jesperpedersen 
As a follow-up:
1. Null attributes now look like:
```
pgexporter_pg_replication_slots_temporary{server="primary", slot_name="repl", slot_type="physical", database="n/a", two_phase_at="0/0", invalidation_reason="n/a"} 0
pgexporter_pg_replication_slots_active{server="primary", slot_name="repl", slot_type="physical", database="n/a", two_phase_at="0/0",invalidation_reason="n/a"} 0
```

2. Null metric values are filtered out during this scrape.

we could also expose `pgexporter_null_metric_values_total` as an internal counter (`like query_errors_total`) to track how often NULL metric values are skipped. Would that be useful?
  
My opinion: it’s not really needed. Since this is an internal issue we’re already handling, exposing it might just add noise for users. The debug info should be sufficient to track it if needed.

jesperpedersen:
@Abdelrhmansersawy 

1) :+1: 
2) No; you are right - we are not doing that

jesperpedersen:
Merged.

jesperpedersen:
@Abdelrhmansersawy Please, cross-port

Abdelrhmansersawy:
@jesperpedersen 
I took a look at pgmoneta and pgagroal, and I see no need for cross-porting because:
1. NULL values in sorting with `strcmp` do not occur, as the other project uses a different approach - sorting based on ART metric keys.
2. Empty attribute values do not occur, as all metrics are hard-coded with no queries or PostgreSQL calls.

Let me know if there’s something I missed.

jesperpedersen:
@Abdelrhmansersawy :+1:

Issue #490, opened 2026-03-25
Fix truncation of long string values in JSON/text serialization


sksingh2005:
Cross port :-  [pgmoneta/pgmonteta/1055](https://github.com/pgmoneta/pgmoneta/pull/1055)

PR #493, opened 2026-03-26
[#490] Fix long string truncation in value serialization

sksingh2005:
Fixes - [#490]

Long string values could be truncated in `src/libpgexporter/value.c `because string serialization used a fixed-size `MISC_LENGTH` buffer. 
This change appends string content directly so long values, including SHA512-length strings, are serialized correctly.


jesperpedersen:
Merged.

PR #500, opened 2026-04-01
Add .dockerignore for container builds

fheikens:
The Dockerfiles in contrib/docker/ are built with the repository root as
context. Without a .dockerignore, the entire repository is sent to the Docker
daemon, including files not needed by the cmake build or the runtime image.

Add a .dockerignore that excludes version control, CI configuration, tests,
monitoring tooling, generated metric files, developer tools, and root-level
documentation.

All paths required by the build remain included:
- src/, cmake/, CMakeLists.txt (cmake build)
- doc/ (man page generation)
- extensions/ (YAML query definitions, installed by cmake)
- contrib/docker/pgexporter.conf (runtime config)

The test/ exclusion is safe because cmake already guards it:
if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test").

No change to build output or runtime behavior. Works with both Docker and Podman.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,46 @@
+# Reduce Docker build context for container image builds.
+# The Dockerfiles in contrib/docker/ use COPY . . in the builder stage,
+# so only files needed by cmake and the runtime COPY are required.
+
+# Version control
+.git
+.gitignore
+
+# CI/CD
+.github
+
+# Tests (cmake skips test/ if absent)
+test
+
+# Monitoring and observability tooling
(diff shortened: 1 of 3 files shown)


jesperpedersen:
@fheikens Please, see https://github.com/pgexporter/pgexporter/blob/main/doc/DEVELOPERS.md#authors

jesperpedersen:
@fheikens Add an issue for each pull request, and reference it in the commit message

jesperpedersen:
Merged.

PR #501, opened 2026-04-01
Add HEALTHCHECK to container Dockerfiles

fheikens:
The container Dockerfiles in contrib/docker/ have no HEALTHCHECK instruction.
This means Docker and Podman cannot report whether the pgexporter daemon inside
the container is actually running.

Add a HEALTHCHECK using pgexporter-cli ping, which checks daemon liveness via
the Unix socket. The check runs as the pgexporter user with the container's own
configuration file.

Parameters: interval 10s, timeout 5s, start-period 5s, retries 3.

pgexporter-cli ping checks the daemon process only, not PostgreSQL backend
connectivity or metrics endpoint availability. This is intentional: the
container is healthy when its daemon is running, regardless of backend state.

No change to runtime behavior. Both Alpine and Rocky 9 variants are updated
identically.


jesperpedersen:
Rebased, and merged.

PR #502, opened 2026-04-01
Remove PostgreSQL server from container runtime images

fheikens:
pgexporter implements its own PostgreSQL wire protocol and does not depend
on libpq or any PostgreSQL packages at runtime. The runtime stages currently
install full PostgreSQL server and client packages that are never used.

Remove all PostgreSQL packages from both Alpine and Rocky 9 runtime images.
Remove the PGDG repository from Rocky 9 and the Alpine edge repositories,
which were only present for the PostgreSQL packages. Add explicit runtime
library dependencies (libyaml, zlib, zstd, lz4, bzip2) for Rocky 9 that
were previously pulled as transitive dependencies of the server package.
Remove bzip2-dev from Alpine (dev headers not needed at runtime).

Builder stages are unchanged. No change to runtime behavior.

### DIFF
--- a/contrib/docker/Dockerfile.alpine
+++ b/contrib/docker/Dockerfile.alpine
@@ -64,19 +64,12 @@ RUN cd build && \
 
 FROM alpine:latest
 
-RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories && \
-    echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
-    apk update && \
-    apk add --no-cache \
-    postgresql17 \
-    postgresql17-client \
+RUN apk add --no-cache \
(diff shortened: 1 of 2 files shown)


jesperpedersen:
@fheikens pgexporter doesn't rely on PostgreSQL binaries - it implements a clean-room protocol interaction

jesperpedersen:
@fheikens We don't libpq5

fheikens:
You're right — pgexporter implements its own protocol and does not use libpq. I got the rationale wrong.

I'll amend the PR to also remove libpq5 (Rocky 9), libpq (Alpine), and the PGDG repo since nothing in the runtime image needs it. The rest of the changes (removing postgresql17-server/contrib/client packages, adding the actual runtime library dependencies) are still valid.

jesperpedersen:
Merged.

PR #516, opened 2026-04-20
[#515] Add per-server tls option to disable backend TLS

### LINKED ISSUE #515: No way to disable TLS for backend Postgres connection; SSLRequest sent unconditionally
**Describe the bug**

`pgexporter_server_authenticate()` in `src/libpgexporter/security.c` unconditionally sends a PostgreSQL `SSLRequest` packet to every backend server immediately after TCP connect, and upgrades the connection to TLS whenever the server responds `'S'`. There is no user-visible way to opt out.

The documentation and several deployments suggest setting `tls = off` under a server section (e.g. `[primary]`) disables TLS for that server — but the config parser only recognizes `tls` in the `[pgexporter]` section (which controls whether the `/metrics` HTTP endpoint is served over TLS, not the backend connection). When used under a server section, it is silently logged as `Unknown` and ignored. Example from our logs at startup:

```
pgexporter: Unknown: Section=primary, Key=tls, Value=off
```

Confirmed in [`security.c` @ `pgexporter_server_authenticate`](https://github.com/pgexporter/pgexporter/blob/main/src/libpgexporter/security.c#L1012): the `SSLRequest` write and the `msg->kind == 'S'` branch that calls `SSL_connect()` have no dependency on any per-server config flag, and the `servers[]` struct has no `tls` member.

**Why this matters** — we observed a double-free / heap corruption inside the SSL path on live pods (upgrading SSL after a TCP reset-by-peer):

```
pgexporter[21000]: SSL_ERROR_SSL: Success (-1)
pgexporter[21000]: Reason: internal error
double free or corruption (!prev)
```

Hits every minute once triggered; the process stays up but all scrape workers fail. Since our backends don't require encryption, being able to disable backend TLS would also avoid this SSL-object lifecycle bug entirely. #374 / #375 addressed some SSL leaks but this crash pattern appears to remain.

**To Reproduce**

1. Deploy pgexporter against a Postgres server that advertises SSL availability (e.g. a CloudNativePG cluster with `ssl=on`).
2. Configure the server section as:
   ```
   [primary]
   host = postgres-rw
   port = 5432
   user = pgexporter
   tls = off
   ```
3. Start pgexporter. Observe the startup log: `Unknown: Section=primary, Key=tls, Value=off`.
4. Scrape `/metrics` and observe the backend connection is SSL-encrypted (TLS handshake happens on the wire).

**Expected**

Either (a) an explicit per-server `tls` option that disables the `SSLRequest` send when `false`, or (b) a clear parser error when `tls` is placed under a server section so users aren't misled. `libpq`'s `sslmode=disable` semantics are the natural reference.

**Version**

`main`.

**PostgreSQL**

PostgreSQL 18 (CloudNativePG).

**libev / OpenSSL / OS**

libev 4.x; OpenSSL 3.x; Fedora-minimal container runtime.

**Configuration**
```
[pgexporter]
host = *
metrics = 5002

[primary]
host = postgres-rw
port = 5432
user = pgexporter
tls = off        # silently ignored, SSL handshake still occurs
```

**Proposed fix**

Add a `bool tls` to `struct server`, parse it under server sections in `configuration.c`, and gate the `SSLRequest`/`SSL_connect` block in `pgexporter_server_authenticate` on `config->servers[server].tls`. Default `true` to preserve existing behavior. Document under `doc/CONFIGURATION.md`. Happy to open a PR.

neurodrone:
## Summary

Fixes #515. Adds an explicit per-server `tls` option (default `true`) so that `pgexporter` can connect to a PostgreSQL backend without negotiating TLS.

Before this change, `pgexporter_server_authenticate()` unconditionally sent the PostgreSQL `SSLRequest` packet and upgraded the connection whenever the server responded with `'S'`, and a `tls = off` line under a server section was silently logged as `Unknown`. Operators had no way to opt out of backend TLS which is painful when the server advertises SSL but the deployment doesn't require it, and when bugs in the SSL path (e.g. double-free on reconnect after a TCP reset) need to be worked around while they're being investigated upstream.

## Default behavior

Default is `tls = on`, so existing deployments are unaffected. Every backend still gets an `SSLRequest` and upgrades to TLS when the server advertises it. Operators explicitly add `tls = off` to opt out.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -77,6 +77,7 @@ See a [sample](./etc/pgexporter.conf) configuration for running `pgexporter` on 
 | user | | String | Conditional | PostgreSQL | The user name. Required for `postgresql` type |
 | data_dir | | String | No | PostgreSQL | The location of the data directory |
 | wal_dir | | String | No | PostgreSQL | The location of the WAL directory |
+| tls | `try` | String | No | PostgreSQL | TLS negotiation policy for this server. `off` skips the PostgreSQL `SSLRequest` and connects without TLS. `try` sends the `SSLRequest` and upgrades if the server offers TLS, otherwise proceeds without TLS (preserves previous behavior). `on` sends the `SSLRequest` and fails the connection if the server declines; `on` also requires `tls_ca_file` to be set so the server certificate can be verified |
 | tls_cert_file | | String | No | All | Certificate file for TLS. This file must be owned by either the user running pgexporter or root. Can interpolate environment variables (e.g., `$HOME`) |
 | tls_key_file | | String | No | All | Private key file for TLS. This file must be owned by either the user running pgexporter or root. Additionally permissions must be at least `0640` when owned by root or `0600` otherwise. Can interpolate environment variables (e.g., `$HOME`) |
 | tls_ca_file | | String | No | All | Certificate Authority (CA) file for TLS. This file must be owned by either the user running pgexporter or root. Can interpolate environment variables (e.g., `$HOME`) |
--- a/src/include/pgexporter.h
+++ b/src/include/pgexporter.h
@@ -93,6 +93,10 @@ extern "C" {
 #define SERVER_FIPS_DISABLED         0
 #define SERVER_FIPS_ENABLED          1
 
+#define SERVER_TLS_OFF               0
+#define SERVER_TLS_TRY               1
+#define SERVER_TLS_ON                2
+
 #define AUTH_SUCCESS                 0
 #define AUTH_BAD_PASSWORD            1
 #define AUTH_ERROR                   2
@@ -307,6 +311,7 @@ struct server
    int minor_version;                                      /**< The minor version of the server*/
    int number_of_databases;                                /**< The number of databases */
    int number_of_extensions;                               /**< The number of extensions */
+   int tls_mode;                                           /**< TLS negotiation mode */
    char tls_cert_file[MAX_PATH];                           /**< TLS certificate path */
    char tls_key_file[MAX_PATH];                            /**< TLS key path */
    char tls_ca_file[MAX_PATH];                             /**< TLS CA certificate path */
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,5 @@ Ahmed Mordi <ahmed.m.hamada2003@gmail.com>
 Ashwani Kumar Kamal <ashwanikamal.im421@gmail.com>
 Frank Heikens <frank@elevarq.com>
 Sakshi Aggarwal <sakshiaggarwal2706@gmail.com>
+Vaibhav Bhembre <vbhembre@coreweave.com>
+
--- a/doc/manual/en/04-configuration.md
+++ b/doc/manual/en/04-configuration.md
@@ -76,6 +76,7 @@ See a [sample][sample] configuration for running [**pgexporter**][pgexporter] on
 | user | | String | Conditional | PostgreSQL | The user name. Required for `postgresql` type |
 | data_dir | | String | No | PostgreSQL | The location of the data directory |
 | wal_dir | | String | No | PostgreSQL | The location of the WAL directory |
+| tls | `try` | String | No | PostgreSQL | TLS negotiation policy for this server. `off` skips the PostgreSQL `SSLRequest` and connects without TLS. `try` sends the `SSLRequest` and upgrades if the server offers TLS, otherwise proceeds without TLS (preserves previous behavior). `on` sends the `SSLRequest` and fails the connection if the server declines; `on` also requires `tls_ca_file` to be set so the server certificate can be verified |
 | tls_cert_file | | String | No | All | Certificate file for TLS. This file must be owned by either the user running pgexporter or root. |
 | tls_key_file | | String | No | All | Private key file for TLS. This file must be owned by either the user running pgexporter or root. Additionally permissions must be at least `0640` when owned by root or `0600` otherwise. |
 | tls_ca_file | | String | No | All | Certificate Authority (CA) file for TLS. This file must be owned by either the user running pgexporter or root.  |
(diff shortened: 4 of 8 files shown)


neurodrone:
Also ran `./clang-format.sh`. No diff on the changed files (local clang-format v22; CI uses v21). Added myself to `AUTHORS` per CONTRIBUTING and force-pushed.

jesperpedersen:
@neurodrone You need to update the manual as well

doc/CONFIGURATION.md
```diff
@@ -77,6 +77,7 @@ See a [sample](./etc/pgexporter.conf) configuration for running `pgexporter` on
 | user | | String | Conditional | PostgreSQL | The user name. Required for `postgresql` type |
 | data_dir | | String | No | PostgreSQL | The location of the data directory |
 | wal_dir | | String | No | PostgreSQL | The location of the WAL directory |
+| tls | `on` | Bool | No | PostgreSQL | Negotiate TLS with this server. When `off`, `pgexporter` skips the PostgreSQL `SSLRequest` and connects without TLS, regardless of whether the server advertises SSL |
```
jesperpedersen: Default is off

neurodrone:
Thanks for the quick response and 👀 @jesperpedersen! Pushed the updates to the manual like you suggested.
                                                                                                                                                                                                                                                                                                  
On changing the default, I'd be worried on flipping to `off`. Today `pgexporter` unconditionally negotiates TLS when the server offers it, so every existing install is on an encrypted backend connection. Defaulting `off` would silently downgrade those connections to plaintext on upgrade, with no config change on the operator's side.                                                                                                                                                                                                                                                                 
                                                                      
 The [pgexporter] tls analogy isn't quite parallel. Correct me if I am wrong, but that knob seems to govern the local `/metrics` listener, where default `off` is reasonable. Backend connections to a remote Postgres sit closer to libpq's sslmode, which defaults to prefer ("negotiate if offered") which is what `tls = on` does here.

If you still want it flipped I'll do it, but wanted to really confirm that you wanted this behavior to change.

doc/CONFIGURATION.md
neurodrone: I should have probably commented inline but attempted to address this [in the comment](https://github.com/pgexporter/pgexporter/pull/516#issuecomment-4283950906) outside this thread.

jesperpedersen:
@neurodrone We will have to look deeper - maybe we need a policy setting instead (off, try, on)

neurodrone:
@jesperpedersen Agreed, that's a cleaner model. We can do something like:
                                                                                                                                                                                                                                                                                                  
  - off: skip `SSLRequest` entirely; unencrypted backend.              
  - try: send `SSLRequest`; upgrade if the server offers, proceed without TLS if it declines. Default, preserves current behavior.                                                                                                                                                                 
  - on: send `SSLRequest`; fail the connection if the server declines. 
 
Maps onto libpq's sslmode={disable, prefer, require}. Certificate verification etc. stays the same.

Happy to rework this PR to the 3-state enum if you prefer.

jesperpedersen:
@neurodrone Great ! `on` will fail as well if there no certificates... Looking forward to your patch !

neurodrone:
@jesperpedersen pushed the three-state version up. Default stays `try` (preserves current behavior), `off` skips `SSLRequest` entirely, `on` requires the server to accept TLS and fails otherwise, and also fails fast if `tls_ca_file` isn't set, so we're not upgrading to TLS without server-cert verification.

Added a round-trip test that drives conf_set and conf_get over the management API for all three values plus the invalid-value rejection path. Please let me know if this looks good or you'd want anything else to be changed.

src/include/pgexporter.h
```diff
@@ -307,6 +311,7 @@ struct server
    int minor_version;                                      /**< The minor version of the server*/
    int number_of_databases;                                /**< The number of databases */
    int number_of_extensions;                               /**< The number of extensions */
+   int tls_mode;                                           /**< TLS negotiation mode: SERVER_TLS_OFF | SERVER_TLS_TRY | SERVER_TLS_ON */
```
bassamadnan: THis looks abit off, we dont need the comment on what the `tls_mode` can be

bassamadnan (commented):
The code honestly looks just fine, thanks for working on this.

Do take a look at CI though its failing. Seems like it will be abit painful to debug why its failing, afterall its github ci.

neurodrone:
@bassamadnan Thanks for your 👀 ! I pushed a change to fix the failing test on CI. Ran it locally to confirm before pushing but let's see if it fails again.

src/include/pgexporter.h
neurodrone: Ack. Addressed via https://github.com/pgexporter/pgexporter/pull/516/commits/0b4f3fbf2dd5e57febef71dd0e7e88beda0587e5.

bassamadnan:
You would have to squash all your commits -- See [CONTRIBUTING](https://github.com/pgexporter/pgexporter/blob/main/CONTRIBUTING.md)

neurodrone:
@bassamadnan Done.

neurodrone:
@bassamadnan Anything else you'd want to be updated here?

bassamadnan:
LGTM

jesperpedersen:
@bassamadnan Go ahead
@neurodrone Can you cross port to pgmoneta, pgagroal and pgvictoria ?

neurodrone:
@jesperpedersen I can take a look at the cross-ports when I get some time on my end, no promises on timing though. :) From a quick scan, `pgmoneta` looks like the most direct mirror, `pgagroal` will need some adaptation since it sits on both sides of the handshake as a pooler, and `pgvictoria` didn't surface any obvious equivalent code path so I'd want to dig in. Will file separate PRs in each repo as and when I get to them.

jesperpedersen:
@neurodrone Thanks ! Looking forward to it.

bassamadnan:
Thank you for your contribution ! @neurodrone

Issue #519, opened 2026-04-23
Deque: support sort by value


Ameensakr:
`pgexporter_deque_sort` currently sorts by tag only. Extend it so a deque can
also be sorted by the stored value

### Reference
Same feature landed in pgmoneta: https://github.com/pgmoneta/pgmoneta/issues/845 (PR https://github.com/pgmoneta/pgmoneta/pull/1084).

Ameensakr:
@jesperpedersen 
I could not find any existing test cases for deque.
Should I add a dedicated deque test as part of this feature, or is there another preferred approach for this repository?

jesperpedersen:
@Ameensakr Add all the deque tests

PR #524, opened 2026-04-26
[#519]support deque sort with different data types

Ameensakr:
[#519] resolved 

# Changes
1. Added a comparator for each data type depending on the deque value.
2. Handled cases where there are mixed data types.


Ameensakr:
@jesperpedersen 
PTAL

jesperpedersen:
Merged.

PR #538, opened 2026-05-19
[#535] Added history module skeleton & config properties

### LINKED ISSUE #535: Add history module skeleton and config properties
#### Goals
 - Add the relevant configuration properties for history of metrics. For ex. `history_interval`, `history_backend` etc.
 - Add structs and file stubs for history (only sqlite for now). ex `history_sqlite.c`, `history.c`

sommye-ctr:
Resolves #535 

#### Changes
- Added config properties for history
- Added basic skeleton functions


jesperpedersen:
@sommye-ctr Lets start with the documentation. Create a new chapter in the manual

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Configuration
+
+In order to enable history, add the following to `pgexporter.conf`
+
+```ini
+[pgexporter]
+
+history          = 5005
+history_interval = 60s
```
jesperpedersen: Why do we need this ?

sommye-ctr: The `history_interval` is the period of how often we want to scrape+store the records in history. The `history_retention` is for how old records are kept before removing.

sommye-ctr: The `history_interval` should have a non zero default value as 0s value doesnt make sense (no snapshot is ever stored).

jesperpedersen: @sommye-ctr I think the key point here is to explain that if there already have been a scrape within this period nothing happens. Any scrape from the "outside" should trigger the store

jesperpedersen: @sommye-ctr A value of 0 means that it is only "outside" scrapes - if enabled

jesperpedersen: @sommye-ctr Which leads to a discussion of what the default should be

sommye-ctr: > I think the key point here is to explain that if there already have been a scrape within this period nothing happens. Any scrape from the "outside" should trigger the store

So basically, scrape can be both from outside (/metrics fetch) and from our history timer. Whenever there is a scrape it is accompanied by a store operation (if history enabled). The user can set `history_interval` which will control how often does our automatic history timer runs. 
For eg. lets say if our `history_interval=60s` so at t=60 one scrape+store is done. Now an outside scrape happens at t=90 so now will the timer reset and run at t=150 (90+60) or it will simply abort the t=120 trigger and now do it at t=180?

> A value of 0 means that it is only "outside" scrapes - if enabled

Hmm that does make sense.

jesperpedersen: I think we should just keep the last store time and use that to check if the job should trigger. We shouldn't overload the servers with requests if unnecessary. So, t=60, t=90, t=180.

Remember this ties into the bridge too

sommye-ctr: @jesperpedersen The bridge do we need to store its values into the db too? or do we keep an in-memory store for it? I am a little confused with the bridge...

jesperpedersen: Well, it is a separate database for sure. The bridge instance will likely not pull any PostgreSQL nodes, but only collect data from pgexporter nodes.

However, as developers we need to test both

sommye-ctr: So maybe we make the `history_path` accept a directory and in it based on where we are getting the data from, make `history.db` and `bridge.db` and store the respective data. As for the history module entirely - it remains generic as it is right now and will be able to handle both normally.

jesperpedersen: Or duplicate the settings with bridge_ prefix

sommye-ctr: Yeah... duplicating the settings would give users more control over separate modules (bridge and prometheus history) at the cost of just some more properties. I'll work on the changes

sommye-ctr:
@bassamadnan @resyfer @jesperpedersen How does this look now?

bassamadnan:
Thanks for the PR , @sommye-ctr , a few questions:
1) what are we keeping the history port for? So that other services know f.eg where backends are stored? (eg sqlite)
2) the code is fine for now but i wanted to know your thoughts on having some virtual functions which every implementation of `history.h`. Heres what I was thinking:
```c
struct history_backend_ops {                                                                                                                                                              
  int (*init)(void);                                                                                                                                                                    
  int (*write_batch)(struct history_record* records, int count);                                                                                                                        
  int (*query_range)(const char* metric, time_t start, time_t end,                                                                                                                      
                     struct history_record** out, int* count_out);                                                                                                                      
  int (*prune)(void);                                                                                                                                                                   
  void (*shutdown)(void);                                                                                                                                                               
};                                                                                                                                                                                        
```
then something like  SQLite registers its implementation, and `history.c` just calls through the function pointers instead of the switch statement. If someone adds a PostgreSQL backend later, or maybe want to push it to a queue even for any external sources, they just fill in a new `struct history_backend_ops` and register it. But obviously, its very new and i was just looking at options; we can worry about this later as well.

bassamadnan:
ofcourse these methods arent final but i think we can ahve discussion on wether we want to continue like this or make a decision later.

sommye-ctr:
> 1. what are we keeping the history port for? So that other services know f.eg where backends are stored? (eg sqlite)

The port is for the rest api. The history will have 2 ways to show the data - the rest api and console. The rest api will be served on the specified port



> 2\. the code is fine for now but i wanted to know your thoughts on having some virtual functions which every implementation of `history.h`. Heres what I was thinking:

Yes, in my proposal I did layout a similar plan but I thought to start the implementation of sqlite by just going normally. Surely the virtual functions approach will be better and ig let's start on this approach from the very beginning - easier for the future work. I'll make the relevant changes

sommye-ctr:
A small note, the `server` in history record will be set to endpoint's `host:port` for the bridge history and for direct history server names will be set

bassamadnan:
Sure @sommye-ctr , LGTM then. I am sorry for the delayed response. My github isn't sending me notifications on my phone.

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,97 @@
...
+
+# History
+
+The history module adds **optional historical storage** so that metric
+snapshots are persisted directly by pgexporter. Once enabled, pgexporter
+periodically forks a *history worker* that takes a snapshot of all configured
+metrics and writes them to a storage backend. A separate *retention worker*
+periodically prunes records older than a configured threshold.
+
+The history data is exposed through a JSON HTTP API served on a dedicated
```
jesperpedersen, line 11: We have to explore this - in relationship to the bridge

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,97 @@
...
+To record a history of your PostgreSQL metrics, add the following to `pgexporter.conf`:
+
+```ini
+[pgexporter]
+
+history           = 5005
+history_interval  = 60s
+history_retention = 30d
+history_backend   = sqlite
+history_path      = /var/lib/pgexporter/history.db
```
jesperpedersen, line 27: I think it is ok for now - but PostgreSQL and TimescaleDB won't have this

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+### Bridge metrics history
+
+To record a history of metrics collected from other pgexporter nodes via the bridge, add:
+
+```ini
+[pgexporter]
+
+bridge_history           = 5006
```
jesperpedersen, line 37: Maybe this is the way forward

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,97 @@
...
+bridge_history_interval  = 60s
+bridge_history_retention = 30d
```
jesperpedersen, line 39: We already have in-mem history for the bridge

doc/manual/en/16-history.md
```diff
@@ -0,0 +1,97 @@
...
+bridge_history_backend   = sqlite
+bridge_history_path      = /var/lib/pgexporter/bridge_history.db
+```
```
jesperpedersen, line 42: So maybe the `history` part is the bridge config

jesperpedersen (changes requested):

jesperpedersen:
@sommye-ctr @bassamadnan I think we need more design discussions in the forum thread before we go on

sommye-ctr:
Lets move this discussion to the forum post then

sommye-ctr:
@jesperpedersen As per the discussion on forum post, the current status of PR matches what is needed rn. So shall we proceed with this?

jesperpedersen:
@bassamadnan PTAL, and merge when ready

bassamadnan:
Thank you for your contribution, @sommye-ctr :)

PR #555, opened 2026-07-01
[#554] Refactored http server side logic

### LINKED ISSUE #554: Refactor http server side logic
Extract the repeated server-side HTTP logic into a shared layer so each service (bridge, prometheus, console) only owns its routes and handlers.

sommye-ctr:
Resolves #554 

### Changes
- Extracted shared HTTP server logic (`http_server.c/h`) - request parsing, TLS probe+handshake, route dispatch, and all response helpers (chunked start/write/end, 400/403/301/200) now live in one place. Each service file keeps only its route table and handlers.
- `main.c` fork boilerplate simplified
- Fixed `pgexporter_http_respond_403()` missing the blank-line header terminator - old response had no `\r\n\r\n` after headers, causing clients to hang.
- Fixed missing terminal chunk in `home_page()` across prometheus and bridge.

### DIFF
--- a/src/include/http_server.h
+++ b/src/include/http_server.h
@@ -1,0 +1,203 @@
+/*
+ * Copyright (C) 2026 The pgexporter community
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list
+ * of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this
+ * list of conditions and the following disclaimer in the documentation and/or other
+ * materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without specific
+ * prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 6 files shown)


sommye-ctr:
One doubt - why are we sending 403 instead of 404 when we didn't find a page (`unknown_page()`)?

sommye-ctr:
Also, the bridge and console have no TLS config fields. Should we add those for both of them?

bassamadnan:
> One doubt - why are we sending 403 instead of 404 when we didn't find a page (`unknown_page()`)?

Should be 404 imo. but which line/function are you tlaking about?

jesperpedersen:
@sommye-ctr Full TLS for all of it, plus 404.

This has to be cross-ported

sommye-ctr:
> > One doubt - why are we sending 403 instead of 404 when we didn't find a page (`unknown_page()`)?
> 
> Should be 404 imo. but which line/function are you tlaking about?

The `unknown_page()` function which was in prometheus and other services (ran when no routes matched)

jesperpedersen:
@sommye-ctr well, the HTTP server need to handle 404s, so maybe the functionality can be moved up

sommye-ctr:
> @sommye-ctr Full TLS for all of it, plus 404.
> 
> This has to be cross-ported

For TLS I think we should handle it in a different PR as we would also need to update the various calls to `prometheus_get()` from console and bridge.

jesperpedersen:
@sommye-ctr That is ok, just remember that this PR has to be cross-ported

sommye-ctr:
> @sommye-ctr well, the HTTP server need to handle 404s, so maybe the functionality can be moved up

Yes the http server handles it and it has been changed to 404 now. 
TLS functionality for console and bridge can be done in a separate PR, the infra for it Is setup though.

sommye-ctr:
Currently, there's this bug too where if the tls is configured for the metrics so the console & bridge would fail as they always go through the HTTP route which would cause redirect and null response data

jesperpedersen:
@sommye-ctr That would fall into 404 as well, I think

sommye-ctr:
> @sommye-ctr That would fall into 404 as well, I think

The main issue is because of the TLS issue I mentioned (it will be fixed in a subsequent PR other than this). According to this PR, when a plan http client hits tls configured port - 301 redirect is responded. Do u want me to change it to 404?

jesperpedersen:
@sommye-ctr No, 301 is ok for now

sommye-ctr:
I ran the orangu review and most of the PR seemed fine (apart from some hallucinated issues). Any feedback from you side @bassamadnan @jesperpedersen

bassamadnan (approved):

jesperpedersen:
@bassamadnan Merge, please

bassamadnan:
Thank you for your work! @sommye-ctr

PR #563, opened 2026-07-23
[#557] Added outbound/inbound TLS support for console & bridge

### LINKED ISSUE #557: TLS support for console & bridge
### Inbound
Add console and bridge `cert/key/CA` to the shared HTTP listener so that console & bridge both support TLS 

### Outbound
Both the console and bridge connects with `secure=false`, if the target is TLS-only the scrape gets a redirect and the console & bridge fail.
The bridge endpoint strips away `http/https` but we need to use it to decide to whether that endpoint requires `secure=true`
Also right now, there is no way to setup outbound CA file (which will be required if the target pgexporter instance has setup a ca file - since `SSL_VERIFY_FAIL_IF_NO_PEER_CERT ` this is also set in security.c)

sommye-ctr:
Resolves #557 

### Changes
- Added TLS support for the console and bridge listeners 
- Added TLS support for outbound scraping, including detecting `https://` endpoints and verifying certs via `scrape_ca_file`
- Added mutual TLS support for scraping via `scrape_cert_file/scrape_key_file`.
- Fixed a bug where the TLS handshake could fail due to a non-blocking socket issue in the forked connection handler.

Ideally the outbound cert/key/ca should be per-endpoint for the bridge but with current way of storing and in most realistic cases having a global `scrape_*` is fine for now.


jesperpedersen:
@sommye-ctr Remember the manual as well

jesperpedersen:
Merged.

PR #566, opened 2026-07-23
feat(tests): add utility tests for string and integer appending

### LINKED ISSUE #565: Add unit tests for utils string helpers
Is your feature request related to a problem? Please describe.

The string-building helpers in src/libpgexporter/utils.c  pgexporter_append, pgexporter_append_int, pgexporter_append_char, and pgexporter_compare_string  have no dedicated unit-test coverage. These helpers are used throughout the codebase to build paths, queries, and log/metric strings, so a regression in any of them is high-impact and easy to introduce unnoticed. There's currently no test_utils.c guarding their behavior.

Describe the solution you'd like

Add an MCTF test file test/testcases/test_utils.c that covers the string helpers:

pgexporter_append appending to a NULL buffer allocates a new string; chained appends concatenate correctly.
pgexporter_append_int / pgexporter_append_char ,produce the expected output, including a negative integer.
pgexporter_compare_string, behaves correctly across NULL/NULL, NULL-vs-non-NULL, equal, and unequal cases.
Since this repo's test/CMakeLists.txt lists test cases explicitly (no glob), the new file will also be registered there.

What is the SQL statement you would like to support

N/A  this is test coverage for internal string utilities, not a query feature.

Additional information

Mirrors the equivalent contribution in pgagroal ([#968]). While writing the tests I found that pgexporter_append(orig, NULL) currently crashes  pgexporter_vappend calls strlen() on each argument unconditionally, so a NULL source dereferences NULL rather than being treated as a no-op (as it is in pgagroal/pgvictoria). I've therefore omitted the "append NULL is a no-op" case from this test and will file that crash separately as a bug so the two concerns stay independent.

Youssef-joe:
docs: update authors list to include new contributor Closes #565

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,3 +37,4 @@ Vaibhav Bhembre <vbhembre@coreweave.com>
 Ameen Sakr <ameensakr623@gmail.com>
 
 Som Shegokar <akashshegokar2186@gmail.com>
+youssef-joe <joe92228@gmail.com>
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Merged.

jesperpedersen:
Welcome to the community !

PR #570, opened 2026-08-01
feat(tests): add unit tests for JSON handling and value creation

### LINKED ISSUE #569: Add unit tests for value, json, and utils modules
## Summary

This issue tracks the addition of unit tests for three modules that currently
have little or no test coverage: `value`, `json`, and `utils`.

## Motivation

While exploring the codebase, I noticed the following public APIs are exercised
by the existing test suite only indirectly (or not at all):

- `value.h` — `pgexporter_value_create`, `pgexporter_value_compare`,
  `pgexporter_value_from_double/float`, `pgexporter_value_destroy`
- `json.h`  — `pgexporter_json_put/get`, `pgexporter_json_append`,
  `pgexporter_json_parse_string`, `pgexporter_json_clone`, `pgexporter_json_clear`
- `utils.h` — `pgexporter_snprintf`, `pgexporter_starts_with/ends_with`,
  `pgexporter_base64_encode/decode`, `pgexporter_is_valid_metric_name`,
  `pgexporter_is_number`, `pgexporter_version_ge`, binary read/write helpers

## Changes

- `test/testcases/test_value.c` -->  new file, 9 tests
- `test/testcases/test_json.c`  --> new file, 11 tests
- `test/testcases/test_utils.c` --> 16 tests appended to existing file
- `test/CMakeLists.txt` --> registers the two new test files

## Testing

All new tests follow the existing MCTF conventions and can be run with the
standard test target.

Youssef-joe:
fixes #569


jesperpedersen:
@Youssef-joe Look at CI
@bassamadnan Can you help ?

bassamadnan:
> @Youssef-joe Look at CI @bassamadnan Can you help ?

cant see the logs :( i think github issue, anyway let me know whatever you need help with @Youssef-joe , do try to build and run as well from your end with ASaN (thats where things start to get messy).

Youssef-joe:
> > @Youssef-joe Look at CI @bassamadnan Can you help ?
> 
> cant see the logs :( i think github issue, anyway let me know whatever you need help with @Youssef-joe , do try to build and run as well from your end with ASaN (thats where things start to get messy).

DONE

src/libpgexporter/utils.c
```diff
@@ -382,15 +382,20 @@ pgexporter_read_int64(void* data)
...
+       * value with the high bit set (e.g. -1).
+       */
+      uint64_t res = ((uint64_t)bytes[0] << 56) |
+                     ((uint64_t)bytes[1] << 48) |
+                     ((uint64_t)bytes[2] << 40) |
+                     ((uint64_t)bytes[3] << 32) |
+                     ((uint64_t)bytes[4] << 24) |
+                     ((uint64_t)bytes[5] << 16) |
+                     ((uint64_t)bytes[6] << 8) |
+                     ((uint64_t)bytes[7]);
```
bassamadnan, line 397: why was this needed? was something failing?

src/libpgexporter/ev.c
```diff
@@ -143,8 +143,8 @@ static int initialize_loop_backend(void);
 /* context globals */
 
 static struct event_loop* loop = NULL;
-static _Atomic(struct signal_watcher*) signal_watchers[PGEXPORTER_NSIG] = {0};
-static _Atomic(signal_cb) signal_callbacks[PGEXPORTER_NSIG] = {0};
+static _Atomic(struct signal_watcher*) signal_watchers[PGEXPORTER_NSIG];
+static _Atomic(signal_cb) signal_callbacks[PGEXPORTER_NSIG];
```
bassamadnan, line 147: why this change? i am abit confused on this one

bassamadnan (changes requested):

src/libpgexporter/utils.c
Youssef-joe: Yupp I think so, as the new test_utils_readwrite_int64 case fails under UBSan. The test round-trips -1LL, which makes all eight bytes 0xFF, and the first term is then (int64_t)255 << 56. That's 255 · 2^56 ≈ 1.84e19, well past INT64_MAX (9.22e18), so it's signed left-shift overflow UB per C11 6.5.7p4.

Reduced repro of the old code:
$ cc -std=c17 -fsanitize=undefined -fno-sanitize-recover=all t.c && ./a.out
runtime error: left shift of 255 by 56 places cannot be represented in type 'int64_t'
It's not theoretical-only for the release build either: this is the little-endian branch, so it's the path taken on x86-64 and aarch64  every platform except the big-endian memcpy fast path above it.

Accumulating in uint64_t makes the shifts well-defined (unsigned overflow wraps by definition), and the final (int64_t) cast is the standard out-of-range conversion implementation-defined pre-C23 but documented as two's-complement wrap by both GCC and Clang, and mandated since C23. Values and behaviour are unchanged; only the UB goes away.

src/libpgexporter/ev.c
Youssef-joe: It doesn't build on Clang. 0 isn't accepted as a null-pointer constant for an _Atomic-qualified pointer type, so {0} is treated as an int-to-pointer conversion in a static initializer:


error: incompatible integer to pointer conversion initializing
       '_Atomic(struct signal_watcher *)' with an expression of type 'int'
       [-Wint-conversion]
error: initializer element is not a compile-time constant
GCC accepts it, which is why CI is green  I hit it on Apple Clang 15, and it bites on upstream Clang 16+ too, where -Wint-conversion became an error by default.

= {NULL} isn't a workaround either: it fixes the struct signal_watcher* array but still fails on _Atomic(signal_cb), since (void*)0 → function pointer isn't a compile-time constant.

Dropping the initializer is semantically identical  these are static-storage objects, so they're zero-initialized before any dynamic init, and all-bits-zero is the null pointer representation on every platform pgexporter targets. signal_pending on the next line keeps its = {0} because sig_atomic_t is an integer type and has no such problem.

jesperpedersen: @Youssef-joe stdatomic is required
@Userfrom1995 Cc:

Youssef-joe: <stdatomic.h> is still there  ev.c:44, and ev.h:46 / pgexporter.h:40 is untouched by this commit. The diff is two lines: it only drops the = {0} initializer. The _Atomic types stay, and all five atomic_store_explicit/atomic_load_explicit sites (1947–1948, 1997–1998, 2042) are unchanged. No atomics were removed.

So correcting one thing I said above: I claimed this bites upstream clang 16+, but I can only reproduce it on Apple clang 15 upstream clang 22 compiles = {0}, = {NULL}, and the bare declaration all fine. Sorry for the bad detail.

So the only question is whether the project wants to accommodate Apple clang 15 at all. If yes, = {(struct signal_watcher*)0} / = {(signal_cb)0} keeps an explicit initializer and compiles clean on both.. I'm happy to learn if I am wrong about that tho.
 If you'd rather not carry a workaround for a deficient toolchain, I'll revert the hunk and drop it from this PR; it's unrelated to the JSON tests anyway.

jesperpedersen: @Youssef-joe clang 15 ? Use homebrew
@Userfrom1995 Double-check, but we are not downgrading our platform due to Mac

Youssef-joe: > @Youssef-joe clang 15 ? Use homebrew @Userfrom1995 Double-check, but we are not downgrading our platform due to Mac

okay, gonna upgrade my Clang, and gonna also try with linux machine

bassamadnan:
Hi @Youssef-joe , do let me know whenver this is ready for review! Thanks for working on this

Youssef-joe:
> Hi @Youssef-joe , do let me know whenver this is ready for review! Thanks for working on this

Okay, let's proceed.. sorry for being that late as I had extremely busy month.. also for no I need more tasks and tips to contribute here and in pgagroal more not just writing tests (while I am okay with writing tests for sure).. but I am eager and passionate of learning more about these projects developing my skills in that area too, and for upcoming days I have more time to invest, that would be great if you can help me for that tho, thx :) 🙏

Youssef-joe:
and for the recent commit; Following up on this thread: net effect of fc22417 + 174a856 is zero.. src/libpgexporter/ev.c on this branch is now byte-identical to main (git diff main...add-json-tests -- src/libpgexporter/ev.c is empty). So there's no code change left to review here; the _Atomic initializer stays exactly as it was.

Per @jesperpedersen's point that we're not downgrading the codebase for an old Mac toolchain, I bumped the project's documented Clang floor to 16+ instead of carrying a workaround: [README.md](vscode-webview://10758i697sha92qkaekn9cg8egu5onaes7khv9hhhl7sl7f3rudp/README.md), [CONTRIBUTING.md](vscode-webview://10758i697sha92qkaekn9cg8egu5onaes7khv9hhhl7sl7f3rudp/CONTRIBUTING.md), and [doc/manual/en/74-building.md](vscode-webview://10758i697sha92qkaekn9cg8egu5onaes7khv9hhhl7sl7f3rudp/doc/manual/en/74-building.md) now say clang 16+ and call out that Apple Clang 15 (Xcode 15's CLT) rejects = {0} on an _Atomic-qualified pointer/function-pointer type. Also switched the macOS CI job to build with Homebrew's Clang (brew install llvm) instead of xcrun -find clang, so CI isn't implicitly pinned to whatever Apple Clang the runner's Xcode ships.

I've upgraded my own local Clang too, so this shouldn't come up again on my end. Good to resolve this thread// let me know if you'd rather see the min-version bump split into its own PR.

Youssef-joe:
@bassamadnan Hello..

jesperpedersen:
@Youssef-joe Don't ping Committers. See DEVELOPERS.md - you need to at least to squash

Youssef-joe:
> @Youssef-joe Don't ping Committers. See DEVELOPERS.md - you need to at least to squash

Okay, done.. I just mentioned him like sending PTAL..

PR #571, opened 2026-08-02
Add socket-decoupled TLS context wrapper

trxvorr:
Drive the OpenSSL state machine through memory BIOs instead of SSL_set_fd, decoupling TLS state from the TCP socket, and route the backend connection path through the wrapper. Includes the serializable TLS 1.3 record layer and MCTF unit tests.

Park/resume is not wired: pgexporter closes backend connections per scrape, so cross-scrape persistence would be a separate opt-in feature.

Ported from pgagroal.


jesperpedersen:
Merged.

Issue #572, opened 2026-08-07
Make metrics scrape respect the history_interval duration


sommye-ctr:
Currently, whenever a metrics scrape happens history is written and therefore the frequency of scrape dictates the history write frequency. 
Only 1 write should happen in each `history_interval` duration

AhmedRamadan433:
Hi Mr. Jesper, can I work on this issue?

sommye-ctr:
I am closing this issue as it has been completed already

PR #573, opened 2026-08-09
[#572] Unified history storage paths into single function

sommye-ctr:
### Changes
- Introduced `pgexporter_history_claim_slot(interval)` / `release_slot()` which are shared by the scrape and periodic tick paths.
- Comparison with `history_interval`, storing of pid and atomic flags done in the function.

### DIFF
--- a/src/include/history.h
+++ b/src/include/history.h
@@ -137,6 +137,25 @@ int
 pgexporter_history_store_metrics(struct prometheus_metrics_container* container);
 
 /**
+ * Claim the exclusive right to write a history snapshot.
+ *
+ * Fails if another writer already holds the slot, or if the current
+ * history_interval bucket already contains a snapshot. On success the caller
+ * must invoke pgexporter_history_release_slot() once it is done.
+ *
+ * @param interval Minimum seconds between snapshots, or -1 to snapshot immediately
+ * @return true if the slot was claimed, otherwise false
+ */
+bool
+pgexporter_history_claim_slot(int interval);
+
+/**
+ * Release a slot claimed by pgexporter_history_claim_slot().
+ */
+void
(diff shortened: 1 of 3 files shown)


jesperpedersen:
@sommye-ctr We have to be careful here - if a user press "Refresh" it should override the interval

sommye-ctr:
> @sommye-ctr We have to be careful here - if a user press "Refresh" it should override the interval

Maybe I am not fully understanding how the history & console should work wrt the interval.
So you're saying that:
- If `history_interval` is set
    - ticker should fire and it should only store if duration since last store is greater than the history_interval
    - The metrics scrape should override any config value whatsoever (I am little confused on this as in one of our prev discussion on mail you wrote "In that case the web console will be read-only...")
- If not set then no ticker fires and the history is stored whenever a metric scrape happens

jesperpedersen:
@sommye-ctr Normally we will gather based on interval, and it is read-only. However, if a user press "Refresh" it has to go through and update at once. We need to have the same in the CLI - the system collects information but the user can "override" with an explicit command - lets say they (or pgvictoria) changed a value so we need the before and after; not waiting for interval to trigger

sommye-ctr:
Hmm okay... Ig then we should remove any interval bound on the "scrape-path" for history and only respect it when going through the "ticker-path". 
Nevertheless, other small changes in the PR are fine.
I'll make the changes to the PR

jesperpedersen:
@sommye-ctr We can have 2 paths to the same core function. One pass the interval, and one pass -1 (meaning right now)

sommye-ctr:
Introduced a common function to handle the atomic vars and the interval checks.

jesperpedersen:
Merged.

PR #575, opened 2026-08-19
[#574] Changed SQLite history schema

### LINKED ISSUE #574: Change sqlite history schema
See https://github.com/pgexporter/pgexporter/discussions/534#discussioncomment-17933228 for more details

sommye-ctr:
Resolves #574
Changes

- Replaced the flat history table with a series/sample split - series holds the metric,
  server and label set once, sample holds one row per series per timestamp
- Added pgexporter_history_label_hash() in history.c - label pairs sorted by key then
  hashed with SHA-256, so a series keeps its identity when the emission order of its
  labels changes.
- Fixed: the server label was pulled out with strstr(labels, "server=\"") which matched
  the tail of keys like upstream_server and stopped at an escaped quote in an earlier
  value. It now goes through the same parser as the hash.

NOTE: Delete the previous db file before running this

### DIFF
--- a/src/include/history.h
+++ b/src/include/history.h
@@ -35,10 +35,20 @@ extern "C" {
 
 #include <pgexporter.h>
 #include <stdbool.h>
+#include <stddef.h>
 #include <stdint.h>
 #include <time.h>
 
 #include <openssl/ssl.h>
+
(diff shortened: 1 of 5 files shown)


jesperpedersen:
@bassamadnan Can you look at this ?

jesperpedersen:
Merged.

PR #576, opened 2026-08-21
Check star_fds allocations in network.c

Pranav-error:
Cross-port of pgmoneta/pgmoneta#1246 (issue pgmoneta/pgmoneta#1245), as requested by @jesperpedersen.

`network.c` contains the same code in all four projects. `star_fds` is overwritten with the result of `malloc` and `realloc` and then passed straight to `memcpy` with no check. `realloc` returns NULL *without* freeing the original, so on failure the original buffer is leaked and then written through a NULL pointer. The `malloc` branch one block above has the same problem.

Allocate into a temporary, check it, and on failure free `new_fds` and `continue` — which is what the loop already does when `bind_host` fails.

## Verification

- `cppcheck --enable=warning` reported `memleakOnRealloc` on this line before the change and reports nothing after.
- `clang-format` reports no changes for the file.
- I have **not** built this project locally, so the change is verified by static analysis and by being character-identical to the pgmoneta fix rather than by a compile here. CI will be the real check.

Added myself to `AUTHORS` and `doc/manual/en/97-acknowledgement.md` per DEVELOPERS.md, since this is my first PR here.

These are OOM-only paths, so practical severity is low.


Pranav-error:
Simplified the malloc branch after review feedback on the pgagroal port (pgagroal/pgagroal#999): `star_fds` is already NULL there, so it assigns directly and checks `star_fds == NULL` rather than going through a temporary. The temporary is kept only in the realloc branch, where `star_fds` still holds the only pointer to the original buffer.

Pranav-error:
Rebased onto current `main`. This one keeps the `AUTHORS` and acknowledgement entry.

While doing it I fixed the reason these kept conflicting: #576, #577, #578 and #579 each added my name to `AUTHORS` and `97-acknowledgement.md` separately, so whichever merged first would make the other three conflict, and they had all gone `CONFLICTING`. The name is now added **only here**; the other three touch source files only. They can merge in any order after this.

Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #587 stays open, so nothing is lost if it is wanted later.

PR #577, opened 2026-08-21
Reposition explicitly before appending to the users file

Pranav-error:
Part of the sweep @jesperpedersen asked for on pgmoneta/pgmoneta#1242. Same change in pgmoneta, pgagroal and pgexporter — `admin.c` has the identical shape in all three.

cppcheck reports `IOWithoutPositioning`: the users file is opened in update mode (`"a+"`), read with `fgets`, then written with `fputs`, with no positioning call in between.

**This is not a live bug today.** C11 7.21.5.3p7 allows input to be followed by output when "the input operation encounters end-of-file", and on the only path that reaches the `fputs` the `fgets` loop has run to completion. Every early exit from that loop is a `goto error`, which skips the write.

But `fgets` also returns NULL on a read error, and in that case the exception does not apply and the behaviour is undefined. Seeking to the end before the write removes the dependence on that exception. The stream is in append mode, so output already goes to the end and behaviour is unchanged either way.

I would rather say that plainly than present a warning-silencing change as a bug fix. If you would rather take an inline suppression instead, that is a reasonable call and I will send that version.

## Verification

`cppcheck` reported `IOWithoutPositioning` in `admin.c` before and reports none after. No `clang-format` changes. I could not configure a build locally, so this is analyzer and formatter only and CI is the real check.


Pranav-error:
Added an AUTHORS/acknowledgement entry, and rebased so this carries only its own commit.

Pranav-error:

Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #588 stays open, so nothing is lost if it is wanted later.

jesperpedersen:
@Pranav-error Why are you closing these ? We will get to them

Pranav-error:
They had been sitting a while, so I took that as them not being wanted and did not want to use up your review time on them. Wrong read on my part — all six are reopened.

jesperpedersen:
@Pranav-error If we don't want it we will let you know and close ourselves

Pranav-error:
Understood — thank you.

PR #578, opened 2026-08-21
Check the allocations used in pointer arithmetic

Pranav-error:
Part of the sweep from pgmoneta/pgmoneta#1242. `cppcheck` reports `nullPointerArithmeticOutOfMemory` — an allocation whose result is used in pointer arithmetic without being checked, so an allocation failure turns into arithmetic on NULL and then a write through it.

`queries.c` — two functions allocate the query buffer with `malloc(size)` and go straight into `memset` and `pgexporter_write_int32(content + 1, ...)` / `pgexporter_write_string(content + 5, ...)`. Both now check before use and take the existing `goto error` path.

## Verification

- `cppcheck` reported these findings before and reports none after.
- `clang-format` reports no changes.
- I was not able to configure a build locally, so this is verified by analyzer and formatter only — CI will be the real check.

These are OOM-only paths, so practical severity is low.


Pranav-error:
Rebased onto current main so this carries only its own commit, and added an AUTHORS/acknowledgement entry.

Also: the commit message previously claimed it guarded the mkdir helper, carried over from the pgmoneta commit this was ported from — but that guard had not actually been applied here. `pgexporter_mkdir()` does `dir + 1` unconditionally, which runs past the terminator for an empty string, and its only caller passes an allocation that can be NULL, so I've added the guard to match what the message says.

Pranav-error:

Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #589 stays open, so nothing is lost if it is wanted later.

PR #580, opened 2026-08-25
Check the hex dump line before measuring it

Pranav-error:
Cross-port of pgmoneta/pgmoneta#1259, merged there as `267e367`.

In the hex-dump branch of `pgexporter_log_mem()`, `l` starts as NULL and is built up one byte at
a time:

```c
char* l = NULL;
size_t count = MIN((int)remaining, (int)LINE_LENGTH);

for (size_t i = 0; i < count; i++)
{
   ...
   l = pgexporter_append(l, &buf[0]);
}

if (strlen(l) == LINE_LENGTH * 2)
```

`pgexporter_append()` forwards to `pgexporter_vappend()`, which returns `orig` unchanged on both
its allocation failure paths (the `strings` malloc and the `realloc`). So if the very first
append fails, `l` is still NULL when `strlen(l)` runs. The loop always executes at least once
— the enclosing `while (remaining > 0)` guarantees `count >= 1` — so there is no path where
`strlen` is reached with `l` untouched for any other reason.

This needs a real allocation failure to trigger, same as the merged pgmoneta#1246 class.

The guard frees `n` and breaks. `break` is enough rather than a new label: `t` is still NULL at
that point, and the `atomic_store(&config->log_lock, STATE_FREE)` that releases the log
lock sits after the loop, so it still runs.

The `int` → `size_t` change on `chars_missing` is the other half of the merged pgmoneta commit;
the subtraction is `size_t` arithmetic and the branch already proves it is positive.

## Verification

cppcheck 2.21.0 reports `nullPointer` at this line before the change and nothing after
(`--enable=warning --inline-suppr`). I could not build here, so the reachability argument is
from reading, not from running it. `clang-format` reports no changes.

Independent of my other open PRs — merge order does not matter.


Pranav-error:
Deliberately no AUTHORS/acknowledgement change here — #577 already adds those two lines and is still open, so adding them again would collide. This PR is otherwise self-contained and can merge in any order.

Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #591 stays open, so nothing is lost if it is wanted later.

PR #581, opened 2026-08-26
Check the realloc result in pgexporter_memory_dynamic_append

Pranav-error:
Cross-port of the same guard that is already in pgmoneta, and of pgagroal/pgagroal#1013 (merged today).

`pgexporter_memory_dynamic_append()` reallocs and then writes into the result without checking it:

```c
s = orig_size + append_size;
d = realloc(orig, s);
memcpy(d + orig_size, append, append_size);
```

On allocation failure `realloc` returns NULL and leaves `orig` valid, so this writes to `NULL + orig_size`. `pgmoneta_memory_dynamic_append()` — the same function in the sibling project — already returns `orig` with `*new_size` left at `orig_size`, which keeps the buffer and its recorded size consistent. This makes pgexporter match it exactly.

**Scope, stated honestly:** `pgexporter_memory_dynamic_append()` has **no in-tree callers** today — it is only declared in `src/include/memory.h` and defined here. So this is not fixing an observable crash; it is bringing an exported function in line with the contract its callers would reasonably expect, and with the three sibling projects. Say the word if you would rather leave unused code alone.

## One related site I deliberately did not touch

`data_append()` in `queries.c:893` has the same `realloc` + `memcpy` shape, but it cannot take the same fix. Both of its callers do:

```c
data = data_append(data, data_size, msg->data, msg->length);
data_size += msg->length;
```

They advance `data_size` unconditionally, so returning `orig` on failure would leave a buffer smaller than `data_size` claims and turn an allocation failure into an overrun later. Fixing it properly needs a caller-side decision about how to abort the read loop, which felt like a separate change rather than something to slip into this one. Happy to send it if you want it.

## Verification

cppcheck 2.21.0 reports nothing at this line either way — I found it by grepping every `realloc(` site for a missing NULL check, not with a tool. No build here, so this is verified by reading; the change is the same three lines already running in pgmoneta. `clang-format` reports no changes.


Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #592 stays open, so nothing is lost if it is wanted later.

PR #582, opened 2026-08-29
Check the reallocs that grow the YAML parser arrays

Pranav-error:
The three places that grow the YAML parser's arrays assign the `realloc` result straight back over the only pointer to the old buffer:

```c
*metrics = realloc(*metrics, (*n_metrics + 1) * sizeof(yaml_metric_t));
memset(&(*metrics)[*n_metrics], 0, sizeof(yaml_metric_t));
```

On failure that does two things at once. `realloc` returns NULL without freeing, so overwriting `*metrics` **leaks the array built so far**, and then `memset` writes through the NULL. The same two lines appear in `parse_metrics`, `parse_queries` and `parse_columns`.

This is why I left these out of #581, where I said the same shape in `data_append` needed a caller-side decision — here it does not, because each of the three functions already has an `error:` label that unwinds properly. So the fix is to realloc into a temporary, and `goto error` when it fails.

## Verification

`clang -fsyntax-only -I src/include` is clean, and `clang-format` reports no changes. I could not run the parser against a malformed YAML file to force an allocation failure, so this is verified by reading plus a syntax check rather than at runtime.

Related, and deliberately not in this PR: `data_append()` in `queries.c` still has the unchecked `realloc` from #581's description. Its two callers advance `data_size` unconditionally, so it needs a decision about how to abort the read loop rather than a guard.


Pranav-error:
Closing this to keep the review queue short — this and the other hardening patches are better sent as a single change than as six separate pull requests. Issue #593 stays open, so nothing is lost if it is wanted later.

PR #585, opened 2026-09-03
Drain the kqueue accept backlog and make the listener non-blocking

### LINKED ISSUE #564: Cross-port pgagroal#923/#952: drain kqueue accept backlog + non-blocking listener in ev.c
### Summary

Cross-port of pgagroal's kqueue accept-stall fix (pgagroal#923 / PR pgagroal#952) into pgexporter's `src/libpgexporter/ev.c`, which shares the same event-loop lineage. Coordinated with the pgagroal maintainers — this is the `ev.c`-sharing project called out for the cross-port.

### The latent bug in `src/libpgexporter/ev.c`

Two related issues, the same pair fixed upstream:

1. **kqueue accept backlog is not drained.** `ev_kqueue_io_handler()` (MAIN case) calls `accept()` exactly once per notification, but the listener is registered with `EV_CLEAR` (edge-triggered). When more than one connection is queued on the backlog behind a single readable edge, only one is accepted; the rest are stranded until the next new connection produces a fresh edge. epoll is level-triggered (`EPOLLIN`) and re-fires until drained, so only the kqueue (macOS/*BSD) path is affected.

2. **Listener is left blocking.** `pgexporter_event_accept_init()` stores `listen_fd` without setting `O_NONBLOCK`. So once a drain loop is added, the terminating `accept()` (which must return `EAGAIN`/`EWOULDBLOCK` on an empty queue) would instead block the main process. The listener must be non-blocking for the drain to terminate.

### Proposed fix (mirrors pgagroal#952)

- Drain the accept queue in a `while` loop in the kqueue handler until `accept()` returns `EAGAIN`/`EWOULDBLOCK`.
- Set the listener non-blocking in `pgexporter_event_accept_init()`, the single point every accepting socket is registered through. This is also correct for epoll (already treats `EAGAIN` as "no more pending") and harmless to the io_uring multishot-accept path.

### Impact

pgexporter runs at far lower connection concurrency than a pooler, so this is unlikely to manifest under normal load — but it is the same latent correctness bug in shared code, worth keeping the projects in sync.

### Reference

- Upstream fix: pgagroal PR #952 (single commit `b74236f`), issue #923.
- The PR will be opened here once #952 merges, to port the exact landed form.

Pranav-error:
Closes #564.

Cross-port of pgagroal/pgagroal#952 (for pgagroal/pgagroal#923) into `libpgexporter/ev.c`, which shares the same event-loop lineage. @fheikens raised the issue here on 2026-07-23 and it has been unassigned since, so I picked it up — happy to hand it back if he would rather carry it.

Two halves, matching upstream:

**Drain the accept backlog.** The listener is registered `EV_CLEAR`, so kqueue reports readiness once per readable transition. `ev_kqueue_io_handler()` accepted exactly one connection per notification, so anything else queued behind that single edge stayed in the backlog until a new connection produced a fresh edge. epoll is level-triggered and re-fires until drained, so only the kqueue path (macOS/*BSD) is affected.

**Make the listener non-blocking.** The drain loop stops on `EAGAIN`/`EWOULDBLOCK`; on a blocking listener that final `accept()` would block the daemon once the backlog empties. `pgexporter_event_accept_init()` did not set it.

One difference from the upstream patch worth flagging: pgexporter's helper takes a second argument — `pgexporter_socket_nonblocking(int fd, bool value)` — where pgagroal's takes only the fd, so this calls it as `pgexporter_socket_nonblocking(listen_fd, true)`.

## Verification

`clang -fsyntax-only -I src/include` is clean and `clang-format` reports no changes. I have not reproduced the stall: it needs several near-simultaneous connections against a kqueue build, and the equivalent change is already merged and running in pgagroal. So this is verified by matching the upstream fix and by reading, not by triggering the original behaviour.


Pranav-error:
Thanks — the first point was right and is fixed in `c9de804`.

`pgexporter_socket_nonblocking()` did ignore both `fcntl` results and always returned 0, so a failure would have been silent. That matters here specifically, because the drain loop stops on `EAGAIN` and would otherwise block in `accept()` on a listener that never became non-blocking. It now checks `F_GETFL` and `F_SETFL`, logs a warning and returns 1 — the same shape as `pgagroal_socket_nonblocking()`, keeping the extra `bool value` argument this version has. `pgexporter_event_accept_init()` now fails with `PGEXPORTER_EVENT_RC_ERROR` rather than continuing.

On the second point — exiting the drain loop when the callback breaks the event loop — I have left it alone for now, deliberately. The upstream fix in pgagroal/pgagroal#952 does not do it either, and the value of this PR is that it matches a patch already merged and running there; diverging from it is a change I would rather have a maintainer choose than make on my own. It is a real consideration though: `MANAGEMENT_SHUTDOWN` can break the loop from inside the callback, and the loop would keep draining the backlog afterwards.

Happy to add the guard here if you want it, and it would probably belong upstream in pgagroal too rather than only here.

Pranav-error:
Both Copilot points addressed, and I filed #586 for the underlying problems and squashed to a single commit referencing it.

**Drain loop vs. shutdown** — valid, fixed. The loop condition is now `pgexporter_event_loop_is_running()` instead of `while (true)`, so a callback that calls `pgexporter_event_loop_break()` (MANAGEMENT_SHUTDOWN) stops the drain instead of continuing to accept. That is the same check the io_uring path already makes after its callbacks.

**`pgexporter_socket_nonblocking()` always returning 0** — this was already the second half of the PR: `network.c` now checks both `F_GETFL` and `F_SETFL` and returns 1 on either failure, matching `pgagroal_socket_nonblocking()`. The comment was reading the pre-PR code.

`clang-format` reports no changes.

PR #595, opened 2026-09-07
Cross-port: Build the number strings with format_and_append

### LINKED ISSUE #594: append_int and append_ulong truncate the largest value
`pgexporter_append_int()` and `pgexporter_append_ulong()` in `utils.c` pass the digit count where `snprintf` wants the buffer size, so each is one byte short and drops the last digit of the largest value of its type:

```c
char number[12];
pgexporter_snprintf(&number[0], 11, "%d", i);

char number[21];
pgexporter_snprintf(&number[0], 20, "%lu", l);
```

`INT_MIN` is `-2147483648`, 11 characters plus a terminator, and `ULONG_MAX` is 20 digits plus a terminator, so:

```
append_int(INT_MIN)      size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

There is no overflow — `snprintf` truncates safely. The result is a wrong number, not a crash, and the practical reachability is low: the current callers pass ports, versions and counters that will not reach those values. It is a trap for a future caller rather than a live bug.

Cross-port of pgagroal/pgagroal#1027, which @jesperpedersen asked to have carried across. The fix there follows his suggestion of using the `_append_` functions: the three helpers become one-line calls to `pgexporter_format_and_append()`, which sizes its buffer from `vsnprintf(NULL, 0, ...)` and so cannot truncate at all. That removes the fixed buffers rather than correcting their sizes.

`pgexporter_append_ullong()` does not exist here, so only the two functions are affected.

Pranav-error:
> **This is a cross-port, not a new proposal.** The original is pgagroal/pgagroal#1028, already reviewed and **merged**. The change here is the same apart from the project prefix, so it should only need a committer's eye rather than a full review.

Fixes #594.

Cross-port of pgagroal/pgagroal#1028, which @jesperpedersen asked to have carried across on pgagroal/pgagroal#1027.

`pgexporter_append_int()` and `pgexporter_append_ulong()` pass the digit count where `snprintf` wants the buffer size, so each is a byte short and drops the last digit of the largest value of its type:

```
append_int(INT_MIN)      size 11 -> "-214748364"            correct: -2147483648
append_ulong(ULONG_MAX)  size 20 -> "1844674407370955161"   correct: 18446744073709551615
```

Rather than correcting the two sizes, this follows the shape jesperpedersen asked for on the pgagroal PR: both helpers become one-line calls to `pgexporter_format_and_append()`, which sizes its buffer from `vsnprintf(NULL, 0, ...)` and therefore cannot truncate for any value. The fixed buffers go away entirely.

`pgexporter_append_ullong()` does not exist here, so only two functions change.

There is no overflow in the current code — `snprintf` truncates safely — so this produces a wrong number rather than a crash, and the reachable impact today is small. I would rather say that than present it as more than it is.

## Verification

Ran the rewritten helpers with `format_and_append` and `append` copied verbatim:

```
append_int(INT_MIN)      = "-2147483648"           len=11
append_int(INT_MAX)      = "2147483647"
append_int(0)            = "0"
append_ulong(ULONG_MAX)  = "18446744073709551615"  len=20
append_int(append_int(NULL, 42), 7) = "427"
```

The last line is the regression check — appending onto an existing string still concatenates.

`clang-format` reports no changes. `clang -fsyntax-only -I src/include` reports the same count of pre-existing errors before and after the change (build-generated defines this environment does not supply), compared against the upstream file rather than a stash, so the count is a real before/after.

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -457,6 +457,34 @@ char*
 pgexporter_append_ulong(char* orig, unsigned long l);
 
 /**
+ * Append an unsigned long long
+ * @param orig The original string
+ * @param l The long
+ * @return The resulting string
+ */
+char*
+pgexporter_append_ullong(char* orig, unsigned long long l);
+
+/**
+ * Append a double
+ * @param orig The original string
+ * @param d The double
+ * @return The resulting string
+ */
+char*
+pgexporter_append_double(char* orig, double d);
+
+/**
+ * Append a double with set precision
+ * @param orig The original string
+ * @param d The double
+ * @param precision The number of digits after decimal
+ * @return The resulting string
+ */
+char*
+pgexporter_append_double_precision(char* orig, double d, int precision);
+
+/**
  * Append a bool
  * @param orig The original string
  * @param b The bool
@@ -473,6 +501,17 @@ pgexporter_append_bool(char* orig, bool b);
  */
 char*
 pgexporter_append_char(char* orig, char c);
+
+/**
+ * Append bytes with an explicit length
+ * @param orig The original string
+ * @param s The bytes
+ * @param s_length The number of bytes
+ * @param orig_length The length of the original string
+ * @return The resulting string
+ */
+char*
+pgexporter_append_bytes(char* orig, const char* s, size_t s_length, size_t orig_length);
 
 /**
  * Indent a string
--- a/src/libpgexporter/utils.c
+++ b/src/libpgexporter/utils.c
@@ -898,17 +898,30 @@ char*
 pgexporter_format_and_append(char* buf, char* format, ...)
 {
    va_list args;
-   va_start(args, format);
+   int len;
+   char* formatted_str = NULL;
 
    // Determine the required buffer size
-   int size_needed = vsnprintf(NULL, 0, format, args) + 1;
+   va_start(args, format);
+   len = vsnprintf(NULL, 0, format, args);
    va_end(args);
 
+   // Leave buf as it is on failure, like pgexporter_append() does
+   if (len < 0)
+   {
+      return buf;
+   }
+
    // Allocate buffer to hold the formatted string
-   char* formatted_str = malloc(size_needed);
+   formatted_str = malloc((size_t)len + 1);
+
+   if (formatted_str == NULL)
+   {
+      return buf;
+   }
 
    va_start(args, format);
-   vsnprintf(formatted_str, size_needed, format, args);
+   vsnprintf(formatted_str, (size_t)len + 1, format, args);
    va_end(args);
 
    buf = pgexporter_append(buf, formatted_str);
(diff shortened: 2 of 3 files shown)


Pranav-error:
Hardened `format_and_append()` in the same change, so the review point is addressed rather than deferred.

This was my mistake: I spotted the unchecked `malloc` while writing the pgagroal PR and only mentioned it there instead of fixing it. Routing more callers through the function is exactly when it should have been dealt with.

```c
va_start(args, format);
len = vsnprintf(NULL, 0, format, args);
va_end(args);

// Leave buf as it is on failure, like _append() does
if (len < 0)
{
   return buf;
}

formatted_str = malloc((size_t)len + 1);

if (formatted_str == NULL)
{
   return buf;
}
```

Returning `buf` untouched on either failure matches what `_append()` already does for its own allocation failure, so callers see no change in contract. It also fixes a second thing the old code had: `size_needed` was `vsnprintf(...) + 1`, so a negative return became `0`, and `malloc(0)` followed by use of the result was undefined rather than detected.

Driven through both paths with the function copied verbatim and `malloc` forced to fail:

```
normal   : "-2147483648" len=11
appended : "-214748364818446744073709551615"
on OOM   : "-214748364818446744073709551615"  (pointer unchanged, no crash)
```

`clang -fsyntax-only` and `clang-format` clean; pre-existing error counts unchanged against the parent commit.

Pranav-error:
Added `append_ullong()` here too, so all four projects expose the same three helpers.

```c
char*
pgexporter_append_ullong(char* orig, unsigned long long l)
{
   return pgexporter_format_and_append(orig, "%llu", l);
}
```

Declared in `utils.h` next to `append_ulong` with the same doc-comment shape as pgagroal's. It has no callers in this project yet — it is there for parity, matching what pgagroal already has.

`clang -fsyntax-only` and `clang-format` clean on both files; pre-existing error count unchanged against the parent commit.

Pranav-error:
Added corner-case tests for the append helpers, per @jesperpedersen's suggestion to attack the core functions with test cases that expose the bug.

These helpers were used throughout the suite but never asserted on, so nothing pinned their behaviour.

The red-before-green proof was done in pgmoneta, where I can build and run the suite (pgmoneta/pgmoneta#1282). Same code, same tests:

```
                            upstream utils.c   with the fix
test_utils_append_numbers        FAIL             PASS      "append_int truncated INT_MIN"
test_utils_append_double         FAIL             PASS      "append_double truncated 1e19"
```

Here the tests compile and the project builds clean, but I have not run the suite locally, so CI is the real check.

There is no `append_double` in pgexporter, so this covers `append_int`, `append_ulong` and `append_ullong` only.

jesperpedersen:
@Pranav-error All projects should be the same within this area

Pranav-error:
@jesperpedersen Made this area uniform across the four projects.

pgexporter was missing `append_double` and `append_double_precision`, which pgmoneta and pgvictoria both have. Added both, built on `format_and_append` like the rest, so they cannot truncate — worth noting because in the two projects that already had them, they carried the same `char number[21]` with a size of 20, and there it is a live bug rather than a corner case: `%lf` prints the whole integer part, so any double from 1e19 up loses its decimals.

The four now expose the same set:

| | pgagroal | pgmoneta | pgexporter | pgvictoria |
|---|---|---|---|---|
| append_int / ulong / ullong | yes | yes | yes | yes |
| append_double / _precision | #1030 | yes | **added here** | yes |
| append_bool | #1030 | yes | yes | yes |

pgagroal is the odd one out and I am fixing it separately, since #1028 is already merged there.

One I did not add: `append_bytes` exists only in pgmoneta. It takes an explicit length rather than a NUL-terminated string, so it is a different shape from the rest — tell me if you want it everywhere too and I will port it.

Tests for the new helpers are included. They compile and the project builds clean here, but I cannot run this suite locally; the red-before-green proof for the same code is in pgmoneta/pgmoneta#1282.

jesperpedersen:
@Pranav-error We need everything in all projects

Pranav-error:
Understood — added `append_bytes` here and to pgagroal (#1032) and pgvictoria (#62), so the append family is now identical in all four:

| | pgagroal | pgmoneta | pgexporter | pgvictoria |
|---|---|---|---|---|
| append, append_char, append_bool | yes | yes | yes | yes |
| append_int, _ulong, _ullong | yes | yes | yes | yes |
| append_double, _double_precision | #1032 | yes | **here** | #62 |
| append_bytes | #1032 | yes | **here** | #62 |

One I have deliberately not ported yet, with a reason rather than a preference: `append_file_chunk`. Despite the name it is not part of this family — it writes a buffer to a file on disk at an expected offset, and it calls `fopen_secure`, `get_parent_dir`, `mkdir` and `get_files`. Of those, pgagroal has **none**, pgexporter is missing `fopen_secure` and `get_parent_dir`, and pgvictoria is missing `fopen_secure`. So porting it means porting that chain too, which is a much larger change than the string helpers and reaches well outside `utils.c`.

Happy to do it — just confirm you want the dependency chain carried across as well, and I will do it as its own issue and PR per project rather than folding it in here.

Pranav-error:
Flagging a mechanical blocker rather than asking for a review: the workflow runs on this repo need maintainer approval for me, and the current head has not had that, so there is no CI result to look at.

For this PR, `25080dbb` (2026-09-08) is sitting at `action_required`. Two earlier revisions were approved and passed — `402f6246` and `7f8d3f8e` on 2026-09-07 — so the pipeline itself is fine with the change; it is just that the latest revision has not been run. #604 is in the same position with `d7ec85b0`.

No rush on the review itself. It only seemed worth mentioning because the PR looks unverified from the outside when the cause is the approval gate rather than anything in the code.

Locally this builds clean and `clang-format` is quiet; I cannot run the pgexporter suite here, which is why CI matters for this one.

PR #602, opened 2026-09-09
[#572] Make metrics scrape respect the history_interval duration

AhmedRamadan433:
This PR fixes issue #572 in src/libpgexporter/prometheus.c.

Problem:
The history interval was ignored during metrics scrapes because pgexporter_history_claim_slot() was called with -1. This caused history to be written every time metrics were scraped.

Solution:
Pass config->history_interval.ms to pgexporter_history_claim_slot(), converting the value from milliseconds to seconds.

This makes sure history is written only once per configured history interval.


jesperpedersen:
@sommye-ctr @bassamadnan PTAL

sommye-ctr:
I think the issue was resolved ealier through #573. We decided on a decision which can be found in the PR's comments. The issue should have been closed already

sommye-ctr:
@AhmedRamadan433 I have closed the issue - apologies for the mixup. It is advisable to not start working on the issue until assigned. Do you want me to close the PR as well?

AhmedRamadan433:
@sommye-ctr No worries, thanks for clarifying! Yes, you can close the PR too.

PR #614, opened 2026-09-17
[#613] Added parse/bind/exec/sync in queries

### LINKED ISSUE #613: Add parse/bind/execute/sync in queries
Add Parse/Bind/Execute/Sync support so values can be sent as parameters, a way to batch many inserts in one round trip.

sommye-ctr:
Resolves #613

### Changes

- Added extended query protocol support: Parse/Bind/Describe/Execute/Sync/Close message
  builders in `message.c`, and `pgexporter_query_execute_params() `/
  `pgexporter_execute_command_params()` which take an ssl/fd instead of a server index
- Added `pgexporter_pipeline_*` to queue many executions of a prepared statement and send
  them in one round trip, rolled back as a whole if any of them fails
- Added` pgexporter_authenticate_host()` so we can connect to a PostgreSQL instance that
  isn't in the server list, `pgexporter_server_authenticate()`is now a wrapper around it
- Added tests in `test_database.c` for parameters (quotes, NULL, doubles, no rows),
  errors, and pipeline commit/rollback/reuse

Existing metric queries are left on the simple protocol since custom YAML queries can
have multiple statements, which the extended protocol rejects. Parameters and results
are text only for now, and pipeline replies are only read once the whole batch is sent,
so callers should sync every few thousand executions.


sommye-ctr:
@jesperpedersen The build macos job should reference main instead of master

jesperpedersen:
@sommye-ctr Can you fix Mac ?
@bassamadnan PTAL

sommye-ctr:
> @sommye-ctr Can you fix Mac ? @bassamadnan PTAL

Done

